PR #4, opened 2016-11-17
systemctl syntax

genxlee:
This fixes systemctl syntax at documentation.

### DIFF
--- a/install-centos.rst
+++ b/install-centos.rst
@@ -32,39 +32,39 @@ Zammad
 
 ::
 
- sudo service status zammad
- sudo service stop zammad
- sudo service start zammad
- sudo service restart zammad
+ sudo service zammad status
+ sudo service zammad stop
+ sudo service zammad start
+ sudo service zammad restart
 
 Only web application server
 ---------------------------
 
 ::
 
- sudo service status zammad-wew
- sudo service stop zammad-web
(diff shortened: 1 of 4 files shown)


monotek:
Your changes actually break the syntax.

randohinn:
@monotek How's that so? What @genxlee did is exactly right and is how the `service` and `systemctl`commands should be writen. `service NAME ACTION` and `systemctl ACTION NAME`. If you don't believe, read https://www.digitalocean.com/community/tutorials/how-to-use-systemctl-to-manage-systemd-services-and-units and https://linux.die.net/man/8/service

monotek:
Its already like that: 
https://docs.zammad.org/en/latest/install-debian.html

He changes it to the opposite:
https://github.com/zammad/zammad-documentation/pull/4/files#diff-20f2a73764d179784a7cf77965cfa7f9

The Centos  part needs indeed some rework. But as we only supporting Centos 7, which uses systemd, its still wrong.

randohinn:
How about https://zammad.org/documentation/install/install-deb then?

monotek:
Sorry. This is the old docu. 
The docu link on zammad.org will be replaces soon by https://docs.zammad.org/

rkaldung:
@monotek check the links and ping me if I missed one

PR #9, opened 2016-12-05
Add some more information for OTRS migration

cornelinux:
Add missing otrs plugin dependency.

For a rails novice it might be good to know, how to actually start the shell.


monotek:
Thanks! Merged :-)

PR #11, opened 2016-12-16
Update contributing-install-source.rst

patrick-fox:
Filename changed in new develop-branch


patrick-fox:
The document is titled "With Nginx & MySQL"... so there should be a hint to change this to mysql2 as the prerequisites install mysql and not pgsql.

monotek:
You're right. My fault...

PR #15, opened 2017-02-16
no need for procmail when using fetchmail

rkaldung:
It's no problem to use fetchmail without procmail. Tested on a production instance (EL7, Zammad 1.3.x):

image

### DIFF
--- a/channel-fetchmail.rst
+++ b/channel-fetchmail.rst
@@ -26,6 +26,7 @@ Create .fetchmailrc
  su - zammad
  cd ~
  touch .fetchmailrc
+ chmod 0600 .fetchmailrc
 
 
 vi .fetchmailrc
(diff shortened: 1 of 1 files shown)


monotek:
Yes, but there is no presorting by mail header possible with fetchmail only.

Therefore I suggest to put both ways in the docu.

rkaldung:
The topic is piping mails into Zammad. Nor the header is wrong or the content. Please proofread the documentation from a (Zammad) newbie perspective.

monotek:
I'm pretty sure newbies would use the the email channel anyway. I still think it's good to show fetchmail and procmail way. So you can use it as newbie or as advanced user...

thorsteneckel:
@monotek could you please extend the documentation to show both ways? IMHO the order should be:

absolute newbie: uses interface
recommended: fetchmail without procmail (as stated by @rkaldung )
advanced: the procmail way (from before the documentation, the user knows how to handle his mail stuff)

Please let me have a look at it afterwards from my newbie devel point of view. Thanks!

monotek:
Done.
https://github.com/zammad/zammad-documentation/commit/462aa70817ef9eb06065c50a5df062837a49bfe8

PR #17, opened 2017-03-22
added missing -C zammad

nkfilis:
otherwise, the tar.gz will unpack into the wrong directory


monotek:
Thanks for the hint. 
Also changed the order of the commands a bit.
Its fixed now in https://github.com/zammad/zammad-documentation/commit/252327a59ff1e3c6d44965cd8aa8dd4f1ddd9f04

PR #20, opened 2017-06-02
update command run docker container

toanalien:


monotek:
docker container is the new docker syntax.

PR #22, opened 2017-07-05
change MySQL / PostgreSQL in text to fit code

laenan8466:


laenan8466:
change MySQL / PostgreSQL in text to fit code

thorsteneckel:
Hi @laenan8466 - thanks for making the effort to comtribute <3 Sadly I have do decline the pull request since the code says **without** PostgreSQL / MySQL (as the text describes). The original is correct.

This is because of the strange `--without` bundler option we have to use 🙃 

Looking forward to see more PR coming 🤓

laenan8466:
Oh my,
thank you for your clarification. I read that and thought I read it the other way around in the installation doc.

thorsteneckel:
Yeah - happened to me, too. Thanks for your effort!

PR #33, opened 2017-07-26
Installation instructions from source for Debian 9

omacchioni:


monotek:
Thanks for the effort but we like to keep the source install instructions as generic as possible.

We'll update the install instructions for Debian 9 as soon it is available on pkgr / packager.io.
Follow this bug if you interested in progress: https://github.com/crohr/pkgr/issues/128

PR #46, opened 2018-02-27
Add documentation for Kerberos 5

muhammadn:
This feature is not released yet but i had created the documentation on how to configure the Zammad instances to support Kerberos 5 authentication.


thorsteneckel:
@monotek - this feature is not completed yet but we already include it into our documentation? I think this will lead to a lot of questions (on the issue tracker).

monotek:
If it's not ready don't create a pull request.

thorsteneckel:
Agree. Reverted. Please don't create unfinished PRs @muhammadn - and please read before merge @monotek . Thanks to both of you anyway 👍

PR #49, opened 2018-03-20
Update install-centos.rst

dignin:
Corrected the spelling on my change. Pls consider adding this warning.


monotek:
Hi Shwan,

i don't think we need this note as it should be clear that http://localhost is port 80. 
Secondly this note would be needed not only on centos.

Greets from Dresden
André

PR #50, opened 2018-03-29
Trying to install 2.4.0 ...

uli-heller:
Had to adjust the ruby version and the database name within database.yml


thorsteneckel:
Thanks for your contribution!

PR #56, opened 2018-05-14
Ubuntu1804

uli-heller:
With these changes I was able to install zammad on ubuntu1804


monotek:
why should zammad user be able to use apt in sudoers?

PR #57, opened 2018-06-04
PostgreSQL restart effects zammad

benhartwich:
Zammad needs to be restarted everytime you restart postgreSQL or update it in order to work properly.

### DIFF
--- a/install-source.rst
+++ b/install-source.rst
@@ -36,6 +36,8 @@ For PostgreSQL (note, the option says "without ... mysql")
 ::
 
  zammad@shell> bundle install --without test development mysql
+ 
+ Please note: everytime you upgrade or restart PostgreSQL you have to restart all zammad services in order to work properly!
 
 For MySQL (note, the option says "without ... postgres")
 --------------------------------------------------------


monotek:
I doubt this. Need to be tested.

benhartwich:
I´ve made a pull request after submitting this report: https://github.com/zammad/zammad/issues/2051
Your answer is the reason, why I don´t want to participate further. You can delete my pull request.

monotek:
Instead of acting like a stubborn child you could just have linked this issue in the beginning and give some better describtion.

It also seems that you've only tested on Debian 9 (with an unsupported Elasticsearch version 6.x by the way) and that's the reason it has to be tested on all other supported environments, before we change the documentation.

kr4ut:
Just for the record found the same errors today in our Zammad logs. Still need to further debug it but i also had crashing ElasticSearch service which didn't survive the package upgrade from 5.6.9 to 5.6.12 due the incompatible ingest-attachment plugin.

monotek:
fixed in: https://github.com/zammad/zammad/issues/2343
close.

PR #63, opened 2018-09-04
Update Add Translation example

olivergrahl:
It seems that the locale in the example needs to be 'de-de', not just 'de'. Also, running without created_by_id/updatead_by_id returns an error.


MrGeneration:
Thanks for your contribution! :)

PR #71, opened 2018-12-12
free software

txt-file:
free software means that you can run, study, redistribute and improve the software
open source just means that you are allowed to read the code


txt-file:
I'll rewrite the PR.

PR #78, opened 2019-02-28
Improved wording regarding Univention products

ngulden:
I improved the Univention specific wording and also tried to make the text reading more fluently.


MrGeneration:

I changed back the LDAP pre filling pat, as the meaning was kinda off. I updated it so the meaning is clearer now hopefully.

PR #91, opened 2019-08-13
enhancing deletion process for users

MrGeneration:
This commit will allow admins to also remove the avatars of affected users, as you won't need them any longer if you delete the user any way.

### DIFF
--- a/console/dangerzone-for-experts.rst
+++ b/console/dangerzone-for-experts.rst
@@ -70,6 +70,9 @@ The following is the real deal. It will delete all tickets linked to a customer 
     ticket.destroy
   end
 
+  puts "  Removing possible Avatars ..."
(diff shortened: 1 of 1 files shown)


thorsteneckel:
This should be already done:
https://github.com/zammad/zammad/blob/92b38dcb75991b45e0cf578dc266f1cb7c5727f7/app/models/user.rb#L1227-L1243

Which is triggered by a Rails callback before a user record is deleted:
https://github.com/zammad/zammad/blob/92b38dcb75991b45e0cf578dc266f1cb7c5727f7/app/models/user.rb#L34

So Avatars should be already deleted and therefore this addition is not needed.

MrGeneration:
Oh my bad!
I didn't see that, I'll close it without merging then. :-)

PR #94, opened 2019-08-15
Adding declaration of suse compatibility

MrGeneration:
packager.io currently does not support SLES15 which breaks compatibility to SUSE Enterprise 15 (and later) as well as openSUSE LEAP 15.0 (and later).

This commit will add remarks that we can only support SUSE earlier SLES 15, as fixing this is out of our scope.


monotek:
I've already thougt about to completly drop suse support.
Is the old version even maintained anymore?

MrGeneration:
Very good point.
Indeed, openSUSE LEAP 42.x is out of their lifecycle (this june).

However, SLES 12 has a LTSS cycle until 2027:
image

I'd like to keep this for now at least for historical reasons (and because SLES 12 is still in service).
But I wouldn't mind a note regarding dropped support and stuff. :-)

monotek:
Maybe we could also try to find another packing method or service which is a bit better maintained...

MrGeneration:
In the mean time a lot has changed on this Repo.
I'll have to start over...

MrGeneration:
I'll be dropping this PR for now. It no longer has relevance.

PR #95, opened 2019-08-21
adding hints for attribtue translation

MrGeneration:
This commit will help to translate attributes whenever needed.

### DIFF
--- a/console/other-usefull-commands.rst
+++ b/console/other-usefull-commands.rst
@@ -21,6 +21,27 @@ This comes in handy if you e.g. added a new state that you need to translate for
  Translation.create_if_not_exists( :locale => 'de-de', :source => "New", :target => "Neu", format: 'string', created_by_id: 1, updated_by_id: 1 )
 
 
+Translating Attributes
+~~~~~~~~~~~~~~~~~~~~~~
+
+By default Zammad will not translate Attributes. A small exception here is the state dropdown, which has translation enabled by default.
+With the following code you can enable translation, this will translate the attribute display name and the display names of values (if it's a value field).
+For this to work, just replace ``{attribute-name}`` against the name of your Attribute. 
+::
+  
+  attribute = ObjectManager::Attribute.find_by(name: '{attribute-name}')
+  attribute.data_option[:translate] = true		# set this to false to disable translation again
+  attribute.save!
+
+.. Note:: Translating value display names works for the following attribute types:
+  
+  * Boolean
+  * Select
+  * Tree Select
+  
+  If you're translating the display name of e.g. an Integer-Attribute, this works as well!
+
+
 Fill a test system with test data
 ---------------------------------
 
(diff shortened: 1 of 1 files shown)


console/other-usefull-commands.rst
```diff
@@ -21,6 +21,27 @@ This comes in handy if you e.g. added a new state that you need to translate for
  Translation.create_if_not_exists( :locale => 'de-de', :source => "New", :target => "Neu", format: 'string', created_by_id: 1, updated_by_id: 1 )
 
 
+Translating Attributes
+~~~~~~~~~~~~~~~~~~~~~~
+
+By default Zammad will not translate Attributes. A small exception here is the state dropdown, which has translation enabled by default.
```
thorsteneckel: "attributes" instead of "Attributes"

That statement is not true. The following attributes get translated:
```
[2] pry(main)> ObjectManager::Attribute.all.to_a.keep_if{ |a| a.data_option[:translate] }.map{ |a| a.name }
=> ["type", "state_id", "pending_time", "priority_id", "type_id", "internal", "vip", "shared", "domain_assignment", "follow_up_possible", "follow_up_assignment"]
```

We should state instead that it only affects custom attributes.

console/other-usefull-commands.rst
```diff
@@ -21,6 +21,27 @@ This comes in handy if you e.g. added a new state that you need to translate for
...
+With the following code you can enable translation, this will translate the attribute display name and the display names of values (if it's a value field).
```
thorsteneckel: `With the following code you can enable translation, this will translate the attribute display name and the display names of values (if it's a value field).`

`With the following code you can enable translation. This will translate the attribute display name and the display names of values (if it's a value field).`

console/other-usefull-commands.rst
```diff
@@ -21,6 +21,27 @@ This comes in handy if you e.g. added a new state that you need to translate for
  Translation.create_if_not_exists( :locale => 'de-de', :source => "New", :target => "Neu", format: 'string', created_by_id: 1, updated_by_id: 1 )
 
 
+Translating Attributes
```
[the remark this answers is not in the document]
MrGeneration: Heading has been fixed with https://github.com/zammad/zammad-documentation/pull/95/commits/60ec1c2406273be9bd302bdfb0ad13cbdeafc686

console/other-usefull-commands.rst
MrGeneration: Updated the wording (to hopefully meet it better) with commit https://github.com/zammad/zammad-documentation/pull/95/commits/60ec1c2406273be9bd302bdfb0ad13cbdeafc686

console/other-usefull-commands.rst
MrGeneration: Done with commit https://github.com/zammad/zammad-documentation/pull/95/commits/c5e8fc1d477a943a42f4f53327a5e4296584e2b7

PR #96, opened 2019-08-26
Adding searchindex-remark for updating elasticsearch

MrGeneration:
This commit will add a small remark about possible blocked deletion-transactions within single node installations (elasticsearch) which will cause reindexing to fail.


MrGeneration:
Hey Ryan,

can you please have a look if the wording is okay to you?
This commit basically adds hints for users with rather busy Zammad installations to temporary stop Zammad installations. This is basically to ensure the reindexing works without any issues and is just a hint for a ratrher small percentage of user base. :)

rlue:
1. Wouldn't [this page](https://docs.zammad.org/en/latest/install/elasticsearch.html) be a better place for this note? Seems kind of weird to say, "Hey, everything you need to know about upgrading ElasticSearch is on this other page... except this one little thing."
2. **"Zammad-Installation" / "Zammad-Service":** In English, hyphens are not used when joining two nouns together to create a new noun (_e.g.,_ train station, hotel room). Use hyphens only when joining words together to create a new _adjective_ (_e.g.,_ open-source, soft-spoken).
3. Also, don't capitalize "Installation" or "Service".
4. > elasticsearch locks the indexes from deletion

   Not a nitpick, but a genuine question: in a thread from 2017, an Elasticsearch developer said, ["We don't have a concept of a lock in elasticsearch."](https://discuss.elastic.co/t/elasticsearch-document-locking-concept/95924/2) What's this about, exactly? (Just hoping to learn more about elasticsearch.)

Overall, I would suggest this rephrasing:

```rst
Upgrade Notes
-------------

After upgrading, remember to rebuild your search index::

   $ zammad run rake searchindex:rebuild

This step may fail if Zammad is under heavy load:
Elasticsearch locks the indexes from deletion if you're pumping in new data, like receiving a new ticket.
(This only applies to single node deployments, not clusters.)

If it does, try killing Zammad first::

   $ systemctl stop zammad
   $ zammad run rake searchindex:rebuild
   $ systemctl start zammad
```

MrGeneration:
I personally disagree with the location of this phrase.
It imho only affects updating Zammad for fairly big installations that need a reindex (for whatever reason).

This issue mainly appeared on busy installations that have a lot background jobs to work on. In some cases it might happen that Zammad tries to delete the index while elasticsearch is currently writing new information into that indice.

This could cause the reindex to fail as the indice was blocked. I'm not sure if this is real blocking or just some kind of "write protection" on the specific document part to ensure it's not overwritten before it's finished writing down the first request. I'm not too much of an expert in that case. ;(

This is the reason why I think it fits best within the update section. 

---

Above comment is also the reason why I beleive that your re-phrasing is not entirely fitting.
It rather sounds like a searchindex needs to be rebuild after every upgrade of Zammad which is not true. 

I've adapted your other hpyhen and capitalizing notes already. 
The branch again is a bit old and struggles a bit with your earlier refactor, but I managed to adjust it to the current base with as few commits and changes as possible.

rlue:
Ah, my 100% mistake! I didn't look closely enough, and thought this sentence

> If you want to upgrade your elasticsearch installation, please take a look at the **elasticsearch documentation**

linked to the "Install Elasticsearch" page in the _Zammad docs._ Yes, I take back my first comment and totally agree on the original placement of this PR.

---

> for fairly big installations that need a reindex (for whatever reason).

This was another source of confusion for me. I think I understand better now; what do you think about this rephrasing?

```rst
If, for whatever reason, you need to rebuild your search index after upgrading, use::

   $ zammad run rake searchindex:rebuild
   
.. warning:: This step may fail if Zammad is under heavy load:
   Elasticsearch locks the indices from deletion if you're pumping in new data, like receiving a new ticket.
   (This only applies to single-node deployments, not clusters.)
   
   If it does, try killing Zammad first::
   
      $ systemctl stop zammad
      $ zammad run rake searchindex:rebuild
      $ systemctl start zammad
```

install/update.rst
```diff
@@ -134,5 +138,9 @@ Update with DEB
 Updating elasticsearch
 ======================
 
-If you want to upgrade your elasticsearch installation, please take a look at the `elasticsearch documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html>`_ 
+If you want to upgrade your elasticsearch installation, please take a look at the `elasticsearch documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html>`_
 as it will have the most current information for you.
+
+.. note:: If your Zammad installation is quite busy or your ``zammad run rake searchindex:rebuild`` fails, you might want to stop the Zammad-Service: ``systemctl stop zammad``.
+  
+  The Reason behind this is that elasticsearch locks the indexes from deletion if you're pumping in new data (like receiving a new ticket). This only affects elasticsearch single nodes and should not affect clusters.
```
rlue, line 146:
```suggestion
If, for whatever reason, you need to rebuild your search index after upgrading, use::

   $ zammad run rake searchindex:rebuild
   
.. warning:: This step may fail if Zammad is under heavy load:
   Elasticsearch locks the indices from deletion if you're pumping in new data, like receiving a new ticket.
   (This only applies to single-node deployments, not clusters.)
   
   If it does, try killing Zammad first::
   
      $ systemctl stop zammad
      $ zammad run rake searchindex:rebuild
      $ systemctl start zammad
```

MrGeneration: You're right, that's much better!

MrGeneration:
Good to go for me!
Thanks. 💪

PR #101, opened 2019-09-16
Fixing code snipped for making states available to UI

MrGeneration:
Affected documentation part:
https://docs.zammad.org/en/latest/console/working-on-tickets.html#make-new-states-available-to-ui

We currently hold the following code snipped:
```
attribute = ObjectManager::Attribute.get(
  object: 'Ticket',
  name: 'state_id',
)
```

The above actually is wrong and suggest to insert the created states ID which is wrong. The snipped needs to be like this:

```
attribute = ObjectManager::Attribute.get(
  object: 'Ticket',
  name: 'state_name',
)
```

This PR fixes this issue and also fixes a typo / visual thing for the earlier corrected "adding a time picker for pending states".

Martin, could you please check the code snippet-Part? :)
Thank you!

### DIFF
--- a/console/working-on-tickets.rst
+++ b/console/working-on-tickets.rst
@@ -130,13 +130,13 @@ Make new states available to UI
 
 Before being able to use the new states within the WebApp, you need to run the following commands to make them available.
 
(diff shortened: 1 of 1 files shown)


martini:
review ok!

Issue #99, opened 2019-09-11
Add SLA communication hack


MrGeneration:
We need to add further documentation to docs and admin-docs for the following issue:
https://github.com/zammad/zammad/issues/1880

This will help users to address this "issue", if it is one. :)

```
Ticket::Article::Type.find_by(name:'note').update!(communication: true)
```

MrGeneration:
This will be solved with the merge of !103

PR #103, opened 2019-09-16
Add communication-trick for SLA notes

MrGeneration:
This will satisfy issue #99 and tell the user how he can ensure that public notes will be counted as communication and thus affect SLA calculation.


MrGeneration:
A rebase of over 60 files because of commit https://github.com/zammad/zammad-documentation/commit/641f241bb9c882aeb8c4db35cdf1f065feea5e0e would be required. As I think that these changes will bring big issues in terms of overview in approval, I'll start new.

PR #106, opened 2019-10-10
add brief description for zammad migration

promisepreston:


MrGeneration:
First of all, thank you very much for your contribution.

While I agree that a brief introduction / description for the Zammad migration would be great, I also beleive that it's too late in the specific part for OTRS migration. Even more, because you also reference "Zendesk" there, which is on another page.

Right now -in my opinion- the text wouldn't be enough for it's own page (it'd just add complexity).
So the main question is if there's general information that we can put on a parent page to make it more useful.

What do you think?

promisepreston:
I think I understand what you mean. However, the brief explanation will help explain to anyone reading the Migration documentation what it's about on a High level.

I think that by modifying the brief explanation to cater for the OTRS and Zendesk pages individually will be better off, since we don't have a parent page for Migration.

What do you think?

MrGeneration:
We've been talking about this and came to the point, that a brief description is a very good idea.
It should at least contain these information:

* What is a migration?
* Can I import to existing data?
* Can I do a "resync" without changing existing data on the destination?
* What imports are available?

Also, we've came to the point that this should be some kind of landing page - it's a bad idea to put it on e.g. the OTRS page.

----

above comments are somewhat of a help for me to keep the information together, I'll take a look at your commit and will work something out. Please give me some time for that. :-)

promisepreston:
Oh, thank you so much.
I feel glad that you've come to a conclusion that a brief description is a very good idea.

I totally agree with you about the landing page (parent page) for the brief description.
Also, some information specifically relating to Zendesk and OTRS can also be included in their respective pages.

MrGeneration:
Closing this PR.
We'll be reworking the migration parts of our documentation. This will also cover a small brief description of the migration.

PR #107, opened 2019-10-10
added further remarks on backing up as non root-user

MrGeneration:
This has been discussed in https://community.zammad.org/t/db-backup-password/3256
and will help users that don't want to run Backups as root-user.

@thorsteneckel can you please review this?


MrGeneration:
Thanks! Merging now.

PR #108, opened 2019-10-18
extend privacy remarks

MrGeneration:
This extends our privacy information by local storage information. This might help administrators to overview the data retention times.

@thorsteneckel
Can you please take a look and especially check that I got the timings right?
If not, I'll gladly correct them. Also, if you think I missed something out, please let me know!


thorsteneckel:
@rlue can you please have a second look? 👀

PR #116, opened 2019-12-12
Fixed user switch to zammad user (behaviour of 'su' changed in Debian)

dreieinhalb:
- Debian/Ubuntu has replaced 'su' implementation with util-linux
(see: "https://sources.debian.org/src/util-linux/2.33.1-0.1/debian/util-linux.NEWS/")
- from then on 'su' should be used with '-' to get a clean environment

### DIFF
--- a/install-source.rst
+++ b/install-source.rst
@@ -153,8 +153,7 @@ Get Zammad
 
 ::
 
- su zammad
- cd ~
+ su - zammad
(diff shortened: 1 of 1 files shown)


dreieinhalb:
Sorry, I realized I have to make different branches for making pull requests for different fixes. I'm closing this PR and reopen one for each fix, to get it cleaner.

PR #125, opened 2020-01-29
Increase visual separation in Danger Zone code samples

rlue:
Per comments on 641f241 by Marcel Herrguth:

> You merged two code blocks \[in admin/console/dangerzone-for-experts.rst\]
> that were divided for a important reason: The user not to run it
> immediately
>
> (I learned in the community that many people have issues reading and
> that sometimes causes bad things)

This commit uses definition lists to clearly separate the steps of
deleting DB records in the Console guides.

### DIFF
--- a/admin/console/dangerzone-for-experts.rst
+++ b/admin/console/dangerzone-for-experts.rst
@@ -29,43 +29,47 @@ Deleting Customers
 
    As such, the examples below will delete not only the specified customers, but **all tickets associated with them**, as well.
 
-.. code-block:: ruby
+Step 1: Select customers by email address
+   .. code-block:: ruby
 
-   # Select customers by email address
-   >> customers = User.where(email: %w[customer@example.com customer@example.org])
+      >> customers = User.where(email: %w[customer@example.com customer@example.org])
 
-   >> customers = customers.joins(roles: :permissions)
-                           .where(roles: { active: true })
-                           .where(permissions: { name: 'ticket.customer', active: true })
-                           .where.not(id: 1)
+      >> customers = customers.joins(roles: :permissions)
+                              .where(roles: { active: true })
+                              .where(permissions: { name: 'ticket.customer', active: true })
+                              .where.not(id: 1)
 
-   # Preview affected users & tickets
-   >> puts customers.map do |user|
-        "Customer #{user.fullname}/#{user.id} has #{Ticket.where(customer_id: user.id).count} tickets #{Ticket.where(customer_id: user.id).pluck(:number)}"
-      end.join("\n")
+Step 2: Preview affected users & tickets
+   .. code-block:: ruby
 
-   # Proceed with deletion
-   >> customers.find_each do |user|
-        puts %{Preparing deletion of customer "#{user.fullname}" (and #{Ticket.where(customer_id: user.id).count} associated tickets)}
+      >> puts customers.map do |user|
+           "Customer #{user.fullname}/#{user.id} has #{Ticket.where(customer_id: user.id).count} tickets #{Ticket.where(customer_id: user.id).pluck(:number)}"
+         end.join("\n")
 
-        Ticket.where(customer: user).find_each do |ticket|
-          puts "  Deleting ticket ##{ticket.number}..."
-          ticket.destroy
-        end
+Step 3: Proceed with deletion
+   .. code-block:: ruby
 
-        puts "  Removing references for user with email #{user.email}..."
-        ActivityStream.where(created_by_id: user.id).update_all(created_by_id: 1)
-        History.where(created_by_id: user.id).update_all(created_by_id: 1)
-        Ticket::Article.where(created_by_id: user.id).update_all(created_by_id: 1)
-        Ticket::Article.where(updated_by_id: user.id).update_all(updated_by_id: 1)
-        Store.where(created_by_id: user.id).update_all(created_by_id: 1)
-        StatsStore.where(created_by_id: user.id).update_all(created_by_id: 1)
-        Tag.where(created_by_id: user.id).update_all(created_by_id: 1)
-        OnlineNotification.find_by(user_id: user.id)&.destroy!
+      >> customers.find_each do |user|
+           puts %{Preparing deletion of customer "#{user.fullname}" (and #{Ticket.where(customer_id: user.id).count} associated tickets)}
 
-        puts "  Deleting #{user.fullname}..."
-        user.destroy
-      end
+           Ticket.where(customer: user).find_each do |ticket|
+             puts "  Deleting ticket ##{ticket.number}..."
+             ticket.destroy
+           end
+
+           puts "  Removing references for user with email #{user.email}..."
+           ActivityStream.where(created_by_id: user.id).update_all(created_by_id: 1)
+           History.where(created_by_id: user.id).update_all(created_by_id: 1)
+           Ticket::Article.where(created_by_id: user.id).update_all(created_by_id: 1)
+           Ticket::Article.where(updated_by_id: user.id).update_all(updated_by_id: 1)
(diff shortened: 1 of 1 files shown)


admin/console/dangerzone-for-experts.rst
```diff
@@ -29,76 +29,85 @@ Deleting Customers
...
-.. code-block:: ruby
+Step 1: Select customers by email address
+   .. code-block:: ruby
+
+      >> customers = User.where(email: %w[customer@example.com customer@example.org])
+
+      >> customers = customers.joins(roles: :permissions)
+                              .where(roles: { active: true })
+                              .where(permissions: { name: 'ticket.customer', active: true })
+                              .where.not(id: 1)
```
MrGeneration: When copying and executing this code part:

```
customers = User.where(email: %w[example1@domain.de example2@domain.de])
customers = customers.joins(roles: :permissions)
                              .where(roles: { active: true })
                              .where(permissions: { name: 'ticket.customer', active: true })
                              .where.not(id: 1)
```

I receive the following error:

```
2.5.5 :003 >                               .where(roles: { active: true })
Traceback (most recent call last):
SyntaxError ((irb):3: syntax error, unexpected '.')
                              .where(roles: { active: true })
                              ^
2.5.5 :004 >                               .where(permissions: { name: 'ticket.customer', active: true })
Traceback (most recent call last):
SyntaxError ((irb):4: syntax error, unexpected '.')
                              .where(permissions: { name: 't...
                              ^
2.5.5 :005 >                               .where.not(id: 1)
puts customers.map do |user|
Traceback (most recent call last):
SyntaxError ((irb):5: syntax error, unexpected '.')
                              .where.not(id: 1)
                              ^
2.5.5 :006 >
```

This is due to the spacings living inside the code block.
I code solve this with using ` \` on each line end like so:

```
customers = User.where(email: %w[example1@domain.de example2@domain.de])
customers = customers.joins(roles: :permissions) \
                              .where(roles: { active: true }) \
                              .where(permissions: { name: 'ticket.customer', active: true }) \
                              .where.not(id: 1)
```

However: I do not know if this breaks anything or brings other kind of issues.
Could you please check this? :)

admin/console/dangerzone-for-experts.rst
```diff
@@ -29,76 +29,85 @@ Deleting Customers
...
+Step 2: Preview affected users & tickets
+   .. code-block:: ruby
 
-   >> customers = customers.joins(roles: :permissions)
-                           .where(roles: { active: true })
-                           .where(permissions: { name: 'ticket.customer', active: true })
-                           .where.not(id: 1)
+      >> puts customers.map do |user|
+           "Customer #{user.fullname}/#{user.id} has #{Ticket.where(customer_id: user.id).count} tickets #{Ticket.where(customer_id: user.id).pluck(:number)}"
+         end.join("\n")
```
MrGeneration: `.join` seems not the be available within a default Zammad rails console and thus does not work:

```
#<Enumerator:0x000000000b2e1020>
Traceback (most recent call last):
        1: from (irb):19
NoMethodError (undefined method `join' for nil:NilClass)
```

I tried to solve this, however, the console returns these kind of "strings":
```
#<User:0x000000000bbecd40>
#<User:0x000000000bbecc00>
```

This does not return any users / tickets as I expected when running this code (to solve the join issue):
```
2.5.5 :020 > puts customers.map do |user|
2.5.5 :021 >                "Customer #{user.fullname}/#{user.id} has #{Ticket.where(customer_id: user.id).count} tickets #{Ticket.where(customer_id: user.id).pluck(:number)} \n"
2.5.5 :022?>   end
```

the result of the above is this:
```
#<Enumerator:0x000000000bbe3e48>
 => nil
```

admin/console/dangerzone-for-experts.rst
```diff
@@ -29,76 +29,85 @@ Deleting Customers
...
+           StatsStore.where(created_by_id: user.id).update_all(created_by_id: 1)
+           Tag.where(created_by_id: user.id).update_all(created_by_id: 1)
+           OnlineNotification.find_by(user_id: user.id)&.destroy!
 
-        puts "  Deleting #{user.fullname}..."
-        user.destroy
-      end
+           puts "  Deleting #{user.fullname}..."
+           user.destroy
+         end
```
MrGeneration: This block is working as expected 💪

admin/console/dangerzone-for-experts.rst
MrGeneration: I fiddled a bit, this code does work:

```
puts customers.map {|user|
           "Customer #{user.fullname}/#{user.id} has #{Ticket.where(customer_id: user.id).count} tickets #{Ticket.where(customer_id: user.id).pluck(:number)}"
         }.join("\n")
```

admin/console/dangerzone-for-experts.rst
rlue: Sorry for the delay. Yes, this is a problem with the way I wrote the code sample.

Ruby allows you to chain method calls like this in a script, but not in the console. In a script, the interpreter looks ahead, sees the `.where` on the next line, and includes that as part of the first line. The console can't do that, and so when you get to the end of `customers.joins(...)` then it just stops there and evaluates what it gets.

The backslashes will fix this without introducing any other issues. Thanks for looking out.

admin/console/dangerzone-for-experts.rst
rlue: Aha! There's a couple of funny things going on here; let me see if I can elucidate them for you.

1. > `.join` seems not the be available within a default Zammad rails console

   The `#join` method is definitely available, but **we're calling it on the wrong thing**. The traceback/error message contains a crucial hint: "undefined method \`join' for nil:NilClass":

   ```ruby
   # see that #join is defined on arrays
   >> [1, 2, 3].join("\n")
   => "1\n2\n3"

   # but not on nil
   >> nil.join("\n")
   NoMethodError: undefined method `join' for nil:NilClass
   ```

   What this means is that everything leading up to the `.join` in the third line is evaluating to `nil` when it should be evaluating to an array.

2. So what is evaluating to `nil`, and why does switching from `do`/`end` to `{`/`}` fix it??

   ```ruby
   # our code sample, simplified
   puts customers.map { block }.join("\n")

   # what we mean
   puts(customers.map { block }.join("\n"))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ intended console output

   # what the interpreter sees
   (puts(customers.map) { block }).join("\n")
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluates to nil
   ```

I will address this in my next commit, as well. Great looking out. Truly, truly grateful for your sharp eyes.

admin/console/dangerzone-for-experts.rst
rlue: In the end, I discovered a portability/comatibility issue with backslashes, so instead I joined the entire `.joins` / `.where` query onto a single line.

Specifically, the Rails console is built on `irb`, but there's an irb alternative out there called pry (which some folks on the development team use), and pry doesn't handle lines of Ruby that start with a dot.

admin/console/dangerzone-for-experts.rst
rlue: I opted for a combination of your `{`/`}` solution and a heredoc, so that the entire `puts` invocation could be kept on one single line.

rlue:
Okay, Marcel! I'm all done with my set of fixes over here; if you're alright with them, feel free to merge!

MrGeneration:
Wonderfull, thank you very much for your hard work!
I've added the users mail address to the preview output to help administrators during their deletion work to see if they hit all users that were requested for deletion.

E.g. when handling deletion requests on our end, this is exactly what I'm doing - the fullname doesn't help me far enough. No big deal. :)

PR #127, opened 2020-03-19
Add Debian 10 and use headless openjdk jre

xf-:


MrGeneration:
Thank you so much for your patience, time and contribution!
Looking good and thus merging! :)

PR #128, opened 2020-04-09
Little typo

svnr-dvnkln:
`articles_ids` seems to be the wrong method, instead it's `article_ids`


MrGeneration:
Well oh well. Thank you very much for fixing that little typo, of course you're right! :)

PR #131, opened 2020-06-16
Updated supported ES version for Zammad 3.4+

martinseener:
Upon update of our on-premise Zammad 3.3 to 3.4 today, we also updated Elasticsearch from 7.5 to 7.7.1 on Debian 9 and it works perfectly with Zammad 3.4+.


MrGeneration:
Doh! You're right, thanks for the heads up!

PR #132, opened 2020-06-18
Extending environment variable documentation

MrGeneration:
-- low priority as of now --

This PR adds a quite useful environment variable to allow adminsitrators to have more than one session worker doing stuff.

Session workers are dedicated workers for all regards of active user sessions. 
This approach does provide more steam for other scheduler based jobs and potentially doesn't block them.

I also was so keen and added remarks for options that might help on performance tuning and also linked the environment page on hardware information page.

Right now a own tuning page is not yet good enough, because such a comprehensive writeup of  such tips -currently- is out of my time scope.
I'd like to stay this way until we have the ressources. I just don't want to make people believe there are tons of magical tipps when they'll only find like 2 options... :D

### DIFF
--- a/appendix/configure-env-vars.rst
+++ b/appendix/configure-env-vars.rst
@@ -68,7 +68,7 @@ ZAMMAD_WEBSOCKET_PORT
 
    Default: ``6042``
 
-.. note:: Remember to update your nginx config to reflect any changes you make here.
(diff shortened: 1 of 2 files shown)


MrGeneration:
Thank you so much for your hard work.
I was fiddling a bit, but it's more like single words over wording and grammar stuff so I suppose that's okay. :-)

PR #135, opened 2020-07-22
other-usefull-commands.rst: added Channel::EmailParser.process_unprocessable_mails

martinvonwittich:
Found here: https://community.zammad.org/t/cant-create-tickets-per-mail-database-error/1077/2 and sorely missed in the documentation :)

### DIFF
--- a/admin/console/other-usefull-commands.rst
+++ b/admin/console/other-usefull-commands.rst
@@ -13,6 +13,22 @@ The below command will do a manual fetch of mail channels. This will also show e
    >> Channel.fetch
 
 
+Reprocess unprocessable mails
+-----------------------------
+
+When Zammad encounters a mail it cannot parse (e.g. due to a parser bug or a malformed message), it will store the mail in ``tmp/unprocessable_mail/<ID>.eml``,  give up on attempting to parse the mail, and will warn on the monitoring page that there are unprocessed mails.
+
+To force Zammad to reattempt to parse those mails, run the following command:
+
+.. code-block:: ruby
+
+   >> Channel::EmailParser.process_unprocessable_mails
+
+In case of a malformed message (e.g. an invalid email address in one of the header fields), you may need to manually edit the mail before Zammad can process it.
+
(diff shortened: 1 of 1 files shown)


MrGeneration:
Thank you very much for your contribution.
While I agree that this is a useful command, I'm afraid that this comes with a great downside:

Zammad does not parse the mail in terms of email routing.
This means especially in terms of "destination group" Zammad will pick the first group.

If you're unlucky this group may be visible to nobody.
I haven't found a good workaround for this yet, I mean apart from manual parsing like here:
https://admin-docs.zammad.org/en/latest/channels/email/advanced/fetchmail.html

What do you think, should we just mention that or should we look for a better solution?

martinvonwittich:
> Zammad does not parse the mail in terms of email routing.

Hm, but why is that so? Seems like something that just should be fixed ^^

If Zammad requires additional data for routing (for example the name of the mailbox that the mail was downloaded from), this data could be stored in an additional file with the same name but a different extension.

> What do you think, should we just mention that or should we look for a better solution?

Yes, just add it to the docs in the meantime. Even better when this behaviour is documented somewhere, because I haven't seen it mentioned in the community.

I would never have noticed because we're using `X-Zammad-Ticket-Group` headers inserted by our mailserver for routing anyway ;D

MrGeneration:
Hey Martin,

thanks for your feedback.
I created an issue for this: https://github.com/zammad/zammad/issues/3145

If you don't mind I'd like to wait for the outcome of this first, because I currently prefer not to touch the same documentation places dozens of times. This also reduces verification / QA on our end and buys me some time.

If this is unacceptable for you, please let me know.

martinvonwittich:
Well, documenting the current behaviour as-is would have the advantage that it reduces the likelihood that people will fall into this trap, because right now this bug report is probably the only public information about it. On the other hand, creating more work for you is of course the last thing I want to accomplish ^^

MrGeneration (approved):

MrGeneration:
Thank you for your patience and effort. :-)

PR #145, opened 2020-09-10
Improved Performance Tuning.

martini:
Setting WEB_CONCURRENCY and ZAMMAD_SESSION_JOBS_CONCURRENT is not an option if hardware is not upgradable. It's always a option.

Enhanced link to certain settings via #performance-tuning anker.


MrGeneration (approved):
I fixed the linking to the performance tuning section and also improved the session list count to provide unique counters.

PR #147, opened 2020-09-10
Introduce SSO documentation

MrGeneration:
This PR introduces a walkthrough for Kerberos based Single Sign On with Zammad. @martini already handled the technical revision on this documentation.

The documentation currently is missing wording QA (typos and usual german to english mistakes ;) ).
I was very motivated to provide information with as least words possible. Removing more information would lead to crucial information missing.


rlue (approved):
Lots of changes here, as usual. Feel free to take a peek now; I will add explanatory notes and comments tomorrow.

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+.. note:: These directions have been confirmed on Windows Server 2016.
+
+1a. Create a service account
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You may use an existing service account if you have one.
+Admin privileges are not required; a normal user account will do.
+
+Select “This account supports Kerberos AES 256 bit encryption” under
+**Properties > Account > Account options**.
```
rlue, line 121: Any chance we could get a screenshot for this?

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+1b. Register an SPN for Zammad
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. note:: Replace the following placeholders in the command below:
+
+   :``<zammad-host>``:              Zammad FQDN
+   :``<service-acct>``:             Service account logon name
+   :``<service-acct-pwd>``:         Service account password
+   :``<domain>``:                   Windows domain
+   :``<master-domain-controller>``: Master domain controller IP/FQDN
```
rlue, line 132: I used angle brackets (`<` / `>`) instead of curly braces (`{` / `}`) for all of these placeholders because the sample Apache vhost config uses actual curly braces:

```apache
RewriteCond %{LA-U:REMOTE_USER} (.+)
```

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+
+1c. Note the secret key and version number
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The output of the command above contains important data for Step 2e below:
+
+.. code-block:: none
+
+   Using legacy password setting method
+   Failed to set property 'servicePrincipalName' to 'HTTP/<zammad-host>' on Dn 'CN=Zammad Service,DC=<domain>,DC=<tld>': 0x13.
```
rlue, line 153: I replaced all of your `zammadsrv` / `THA` / `DEV` values here with more placeholders for the sake of anonymity.

appendix/single-sign-on.rst
MrGeneration: Sure, good point!

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+Kerberos realm configuration is how you tell the Zammad server
+about the *domain controller* (Active Directory server).
+
+.. note:: Replace the following placeholders in the sample config below:
+
+   :``<domain>``:                   Windows domain
+   :``<domain-controller>``:        Domain controller IP/FQDN(s)
+   :``<master-domain-controller>``: Master domain controller IP/FQDN
```
rlue, line 256: I don't know anything about Windows system administration, so I was really guessing at what these placeholders were supposed to represent.

Did I get it right?

appendix/single-sign-on.rst
MrGeneration: There you go :)

image

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+   [realms]
+           # multiple KDCs ok (one per line)
+           <domain> = {
+                   kdc = <domain-controller>
+                   admin_server = <master-domain-controller>
+                   default_domain = <domain>
+           }
+
+   [domain_realm]
+           .<domain> = <domain>
```
rlue, line 286: I removed a comment here about the leading `.` not being a typo because I wanted to keep the sample configs as clean as possible, but I'm open to putting it back if you feel strongly about it.

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+.. note:: Replace the following placeholders in the commands below:
+
+   :``<zammad-host>``: Zammad FQDN
+   :``<domain>``:      Windows domain
+   :``<secret-key>``:  Secret key (**omit the leading** ``0x``)
+   :``<vno>``:         Secret key version number
+
+   The secret key and version number were found in :ref:`sso-register-spn` above.
+
+.. code-block:: sh
```
rlue, line 307: Using `sh` syntax highlighting here highlights the word `for` in the third line (which is bad). The benefit is that it also highlights the code comments. Not sure how much you care either way; I'm on the fence and would be fine removing all syntax highlighting from this code sample.

appendix/single-sign-on.rst
```diff
@@ -0,0 +1,521 @@
...
+   adds a **Sign in using SSO** button to the sign-in page.
+
+.. note:: 
+   On older versions of Zammad,
+   visit ``https://your.zammad.host/auth/sso`` to sign in.
+
+Step 4: Configure Client System (Windows Only)
+----------------------------------------------
+
+Finally, Zammad users on the Active Directory server’s local intranet
```
rlue, line 426: Is "local intranet" the right word here?

appendix/single-sign-on.rst
MrGeneration: I felt that it may be confusing but if you say it didn't confuse you and you understood why it's there, I'm fine with leaving it as is. :)

appendix/single-sign-on.rst
MrGeneration: Quite on point!
The realm configuration is about how your server is to reach the responsible domain controllers. I think your leading in sentence fits with that.

One thing that concerns me is actually the `admin_server` and `kdc` part.
Originally the file said this:
```
                      # you can define more than one kdc (each on it's own line)	
                      # this allows you to provide secondaries if needed	
                      kdc = {IP / FQDN of domain controller}	
                      # admin_server can be the same as kdc if it's not read only	
                      admin_server = {IP / FQDN of master domain controller}
```

Which actually is a hint on how to configure the file. We may thus want to add it above or below the sample.
Per `kdc` line you may only provide one domain controller. In terms of KDC it doesn't matter if the domain controller you define here is read only (this part is not relevant in my opinion)
The `admin_server` *can* be the same as one of the above mentioned kdcs. This may be true if you e.g. only use one domain controller. Not sure if that helps.

appendix/single-sign-on.rst
MrGeneration: Yes!
It's he actual windows wording for intranet sites.

image

Adding the fqdn of the zammad server (or whatever you use to call Zammads webpage) has to be configured as a local intranet site. This tells the users browser that it can trust the page in question and run the kerberos authentication against that page.

Other wise the browsers won't

appendix/single-sign-on.rst
MrGeneration: Would be fine to me, it's shell content.

appendix/single-sign-on.rst
rlue: Thanks. I do prefer to avoid telling the reader anything whenever possible, because we're already telling them a _lot,_ and each new piece of information adds to the cognitive burden of getting through the guide. I think it also helps that the line is almost duplicated just below:

```
.<domain> = <domain>
<domain> = <domain>
```

which would make it _very_ strange if the dot were a typo.

In any case, we'll know if we need to put it back when we do (or don't) get complaints from the readers, right? :wink:

appendix/single-sign-on.rst
rlue: Okay great.

Technically, a lot of it is the `ktutil` interactive prompt, which is terminal/cli but not exactly the shell, but there's no sense in splitting hairs. :joy:

appendix/single-sign-on.rst
rlue: Got it. Just to make sure I understand you:

1. If you don't take this configuration step, you can still use SSO to sign in, but you have to manually enter your Active Directory username and password.
2. If the Zammad server is actually on the local intranet, you still have to do this.

Are the above points correct?

appendix/single-sign-on.rst
rlue: * “Per `kdc` line you may only provide one domain controller. In terms of KDC it doesn't matter if the domain controller you define here is read only (this part is not relevant in my opinion)”

  Tried to cover that in [this line](https://github.com/zammad/zammad-documentation/pull/147/files#diff-27d6b2138d1c132cf617d7a5301c17a0R282), but happy to rephrase it as:

  ```
  # multiple KDCs ok (one `kdc = ...` statement per line)
  ```

* The `admin_server` can be the same as one of the above mentioned kdcs. This may be true if you e.g. only use one domain controller.

  This is covered in [these lines](https://github.com/zammad/zammad-documentation/pull/147/files#diff-27d6b2138d1c132cf617d7a5301c17a0R262-R263), which appear as part of the definition list item:

  Screenshot_20201002_102807

Thoughts?

appendix/single-sign-on.rst
MrGeneration: Yes, both points are correct.
Sometimes the environments are configured to "detect" the intranet pages automatically, but I wouldn't cover that or rather ignore it..

appendix/single-sign-on.rst
MrGeneration: Okay, let's try it that way then and see what happens.

appendix/single-sign-on.rst
rlue: I've just pushed the following revision:

Screenshot_20201002_222145

This wording okay with you?

MrGeneration: Yep, that's fine to me! Thanks!

PR #156, opened 2021-01-18
Rework Installation documentation

MrGeneration:
This pull request comes with a huge rework on the installation sections.r
It's main goal is to reduce maintaining, duplicate and partly miss leading information.

It also extends the documentation with it's very first "configure webserver" section which should help first timers stepping in easier.

### URL changes

* **removed files without replacements**
   * `/appendix/repo-file.rst`
* **Removed files with replacements / merges**
   * `/install/`
   * `/install/`
   * `/install/`
   * `/install/`


martini (approved):
Good work!

PR #159, opened 2021-02-08
Document user/user[] when using application/json

EtienneBruines:


MrGeneration:
Sorry but could you share more input on the confusion addition?

Because two sentences earlier we state "don't use `user` only" and on your addition for json contexts you're like "user `user`"

That's super confusing.

EtienneBruines:
Yes, sorry. I meant to be short in the docs because there's limited readability.

When making a `HTTP POST` with `Content-Type: application/json`, the following is expected:

```json
{
    'event': 'answer', 
    'from': '0123456789', 
    'to': '150', 
    'direction':  'in', 
    'callId': '32241505-8f8d-4d02-b926-f5a995e4f3ad',
    'answeringNumber': '150', 
    'user': 'Etienne Bruines'
}
```

Using `user[]` does not result in an error, but does not get parsed either. Using `user: ['Name']` resulted in a Server Error.

MrGeneration:
Thank you for the clarification and your patience.
I believe that your point is a very important and valid one.

The issue I currently see with this approach is that the context is missing.
This makes the already confusing part of this documentation worse (just my opinion).

I'd prefer a overhauling the CTI documentation part to meet our current standards and provide a much better experience.
(We're currently doing that on a lot of places - the biggest last overhaul  was done to the API documentation: https://github.com/zammad/zammad-documentation/commit/41984514647457e8da95446e1667abf17f4605c9c - current upcoming would be Backup & Restore: https://github.com/zammad/zammad-documentation/pull/211 )

Since you put some effort into your findings I wouldn't want to "steal" your findings commit wise.
Would you want to give that rework a go or should I go ahead in the near future?

EtienneBruines:
@MrGeneration I probably wouldn't get around to doing it - so feel free to do it :smile:

PR #160, opened 2021-02-11
Intro: Correct HTTP Basic Auth link

jaller94:
* Transforms the Markdown syntax into rst syntax.
* Gives the link a proper name for better accessibility.
  * Automated tools (like screen readers) may list all links on a page and a link with the name "here" doesn't tell you what to expect.


MrGeneration (approved):

MrGeneration:
Thank you very much for your contribution.

PR #176, opened 2021-04-19
Changed default file source

b90g:
As noted in https://github.com/zammad/zammad/issues/3477#issuecomment-822550901 using the FTP file contains the inital release. 

Also i changed the Here Link

Update of #175


MrGeneration (approved):
Thank you very much for your contribution! :)

PR #177, opened 2021-04-22
Updating variables documentation

Azertooth:
adding the nginx server port


MrGeneration (approved):
I adjusted the wording for the variable to reflect the setting being relevant for nginx only better.

PR #181, opened 2021-05-14
package.rst fix centos permissions fix

KeptnArgo:
There seems to be an issue with fix for the CentOS 8 permissions. 
I figured this is how it's supposed to be, not 100% sure about the `chown -R zammad /opt/zammad/public`. 
`find /opt/zammad/public/ -type d -exec chmod +x {} \;` basically sets the +x modifier only for directories, not all files.

### DIFF
--- a/install/package.rst
+++ b/install/package.rst
@@ -195,7 +195,8 @@ Install Zammad
 
          .. code-block:: sh
 
-            chmod -R 755 /opt/zammad/public/
+            chown -R zammad /opt/zammad/public/
+            find /opt/zammad/public/ -type d -exec chmod +x {} \;
 
       .. tab:: OpenSUSE / SLES
 


MrGeneration:
Why are you change ownering the directory first? 
That should be default as the package installation and source code installation documentation do cover that?

Did you verify your steps to be working as expected?

KeptnArgo:
>Why are you change ownering the directory first?

thats the way it was in the original documentation. `chown -R 644` doesn't make much sense to me, not sure if it`s supposed to be `-R zammad` or `-R :zammad`, though the latter one doesn't make much sense to me either. 

>Did you verify your steps to be working as expected?

it's working for me, with the original steps we got some permission errors (can't remember the exact error though)

MrGeneration:
Sorry for the wait.

So I followed the documentation and, except for our commands, ran your proposed commands on CentOS 7 and 8.
The directories to not show what I was expecting.

I then run our command after to show the difference.
I believe your approach is (no longer?) working as intended / expected.

At this state I'd like to leave the current situation as is.

MrGeneration:
Closing due to missing feedback.

PR #189, opened 2021-06-23
 Updated PostgreSQL version requirements to 9.3+ for Rails 6 / Zammad 5.

mgruner:


mgruner:
@MrGeneration I rebased this and had to create a new PR, hope this works.

Please check the repo settings though, I only used `develop` because it is the default branch and you get it on `git clone`. Maybe delete it, also in other repos where it is not used?

MrGeneration:
`develop` is the default branch because we're using the protected branch functionality to enforce Merge Requests and QA. 
We do this to ensure some kind of quality. I have no better solution right now for that.

MrGeneration (approved):

PR #191, opened 2021-07-06
Improve explanation of stable branches.

rolfschmidt:


rolfschmidt:
Hi @MrGeneration ,
@martini agreed to this today. Maybe you remember our discussion to this. Can you QA and merge pls? We should merge this soon.

contributing/branches.rst
```diff
@@ -35,5 +35,7 @@ Stable
 Stable-X.x
 ==========
 
-* There will be several more stable branches because we'll support the last three major versions of Zammad
-* If current stable version is 1.2.0 then the name of the branch is stable-1.2 and there also would be stable-1.1 and stable-1.0
+* Each minor and major version will have a stable branch (stable-x.x).
+* We are supporting the current develop and stable branch for security fixes and minor bugfixes
+* The latest downloadble version on https://zammad.org/ is based on the stable branch
```
MrGeneration, line 40: I don't think that this is relevant. If it is, it's lacking input because https://ftp.zammad.com/ contains much more. 
I don't see the benefit, especially because https://zammad.org/ always links to the release day stable version and is **never** updated on rolling stable. This makes day one patches useless. 

I'd remove line 40 entirely.

contributing/branches.rst
```diff
@@ -35,5 +35,7 @@ Stable
 Stable-X.x
 ==========
 
-* There will be several more stable branches because we'll support the last three major versions of Zammad
-* If current stable version is 1.2.0 then the name of the branch is stable-1.2 and there also would be stable-1.1 and stable-1.0
+* Each minor and major version will have a stable branch (stable-x.x).
+* We are supporting the current develop and stable branch for security fixes and minor bugfixes
```
MrGeneration, line 39: What exactly do we want to tell here? 
We're suddendly talking about `develop` branch even through we're talking about stable Branches...?

What's the goal to do that? 
I wouldn't mention it at all.

I believe you want to say something like this for line 39:

```
We're supporting the current stable branch for security fixes and minor bugfixes.
```

You'll see why I'm going that short for line 41.

contributing/branches.rst
```diff
@@ -35,5 +35,7 @@ Stable
...
+* The latest downloadble version on https://zammad.org/ is based on the stable branch
+* If your version is older and e.g. equals version 1.2.0 then the name of the branch is stable-1.2 and there also would be stable-1.1 and stable-1.0
```
MrGeneration, line 41: That's kinda confusing, I'd stretch it slightly more to give context. 

```
Let's say you're on version `1.2.0` and the next release will be `1.3.0`. 
In that case we'd provide a stable backport of mayor bug fixes that won't interfere with the old code base to `stable-1.2.1` - this also applies for security fixes *if the code base supports it*. The old stable version will be switched to `stable-1.2.0` (from `stable`) and stable version `1.3.0` will be the new `stable` branch.
```

I think that lights the pictures a bit better. 
Wording might be off because it's on a short period.

But I believe that's what you wanted to say and what we should say.

MrGeneration (changes requested):

MrGeneration (approved):

PR #194, opened 2021-07-20
created troubleshooting/faq section and added paragraph about elasticsearch issues

kaeltis:
After having trouble with missing data from the Web-UI and realizing this seems to be a common occurence I was asked to create an addition for the documentation.

I'm hoping this could be extended in the future to add more common issues and their fixes as well as frequently asked questions to reduce opened GitHub issues without creating the same forum threads again and again.

First time using Sphinx / ReStructuredText so I hope I didn't do any major mistakes 😄

Original Issue: https://github.com/zammad/zammad/issues/3662

### DIFF
--- a/appendix/troubleshooting.rst
+++ b/appendix/troubleshooting.rst
@@ -1,0 +1,102 @@
+Troubleshooting and FAQ
+***********************
+
+This guide will discuss frequently asked questions and how to resolve common problems with Zammad.
+
+
+Data missing from the Web-UI
+============================
+
+A commonly reported issue is data missing from the Web-UI.
+This could be Tickets, Articles, Users or anything else :doc:`indexed by Elasticsearch </install/elasticsearch/indexed-attributes>` and can be caused by missing or incomplete indexes.
+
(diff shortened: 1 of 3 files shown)


MrGeneration:
Thanks for your contribution. Due to high load on my end this will take me some time to look over. 
I probably will change the structure slightly to fit our current documentation styles we implement better. 

First look however looked good.

MrGeneration:
Thank you so much for your patience.
I didn't have to rework as much as I originally anticipated because the actual commit was superb already.

I mainly fixed line length and slightly re-arranged the structure.
index might look odd for now, but as we will probably extend the troubleshooting section more and more, we will want to have a index page which leads to all those different steps on a later point. This step actually spares us from a redirect in the docs.

I hope you can still identify with the text. I'm also letting a colleague have a second wording look and then we should be ready to go. About time. :) 💪

YetAnotherGerrit (approved):
Looks good. Nothing to add!

MrGeneration:
The new section will be available within the next 15 - 30mins. 🎉

kaeltis:
Awesome! 🙂

PR #199, opened 2021-09-21
Enhance untar command for easier usage and enhance mysql information

MrGeneration:
This pull request improves the unpack commands for github based tars (they contain a folder we don't want). This reduces load on administrators.

Additionally this PR enhances input on what MySQL servers need to provide to help users choosing that (not favored) way.

### DIFF
--- a/install/source/include-get-the-source.rst
+++ b/install/source/include-get-the-source.rst
@@ -5,6 +5,6 @@ or find the initial version at https://ftp.zammad.com.
 
    $ cd /opt
    $ wget https://github.com/zammad/zammad/archive/stable.tar.gz
-   $ tar -xzf stable.tar.gz -C zammad
+   $ tar -xzf stable.tar.gz --strip-components 1 -C zammad
    $ chown -R zammad:zammad zammad/
    $ rm -f stable.tar.gz
--- a/install/source.rst
+++ b/install/source.rst
@@ -221,6 +221,21 @@ older versions. A list of required versions can be found on the
 
                   $ zypper install libmariadb-devel
 
+      Create database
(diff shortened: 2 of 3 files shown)


install/source/include-get-the-source.rst
```diff
@@ -5,6 +5,6 @@ or find the initial version at https://ftp.zammad.com.
 
    $ cd /opt
    $ wget https://github.com/zammad/zammad/archive/stable.tar.gz
-   $ tar -xzf stable.tar.gz -C zammad
+   $ tar -xzf stable.tar.gz --strip-components 1 -C zammad
```
mgruner, line 8: When I do this on my mac, it fails because the `zammad` directory is missing. Is this due to an outdated `tar`? Or should we include an `mkdir`?

```
❯ tar --version
bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.6
> tar -xzf stable.tar.gz --strip-components 1 -C zammad
tar: could not chdir to 'zammad'
```

mgruner (commented):
Rest looks fine!

install/source/include-get-the-source.rst
MrGeneration: That's not required because we're creating the user before:
https://github.com/zammad/zammad-documentation/blame/a03820859b97b430434bfe14190c901349e50e94/install/source.rst#L45

This automatically adds the required dir.
If you skip that step you'd have to manually add the directory-  however, you'd also run Zammad with a privileged user or in unsupported scope.

mgruner (approved):

Issue #84, opened 2019-07-01
Enhance API-Documentation


MrGeneration:
We should enhance our documentation.
It was noted on a blog that Zammads documentation is "in a bad state".

This especially affects the API part where we should enhance the following:
* [x] add full, real outputs from the API as references.
* [x] explain stuff better
* [x] add more endpoints
* [x] (my approach) tell the user how he can check for more endpoints if we in case missed some that might be important to other people

Reference Blog post was: https://rixx.de/blog/mobile-zammad-notifications-with-pushover/

PR #204, opened 2021-11-03
Add documentation for linking tickets

Nitair:
Hi everyone,

I decided to create another page for linking tickets as I was writing an integration to move from a custom ticket-system to Zammad, therefore I needed to adapt the linking of the tickets as well. After some research in the community forums like https://community.zammad.org/t/link-tickets-as-child-via-api-call/244 or https://community.zammad.org/t/linking-tickets-via-api/3839 which does not contain any answer, I have decided to go further in order to solve my problem. Maybe this documentation will help some people in similar situations.

I'm open for any feedback. 

Just another thing beforehand: At the GET part I have posted a full response (which is quite big). I was struggling if a shortened version would be better. In my opinion the more information the better to work with.

EDIT: While checking https://github.com/zammad/zammad-documentation/issues/84 it seems the full one is better (point: _add full, real outputs from the API as references._)

### DIFF
--- a/api/ticket.rst
+++ b/api/ticket.rst
@@ -14,3 +14,4 @@ For better overview the grouped those up:
       /api/ticket/states
       /api/ticket/tags
       /api/ticket/mentions
+      /api/ticket/tags
--- a/api/ticket/links.rst
+++ b/api/ticket/links.rst
@@ -18,7 +18,7 @@ Request::
 
    GET /api/v1/links
 
-   with following parameters:
+with following parameters:
 
    {
       "link_object": "Ticket", 
--- a/api/ticket/links.rst
+++ b/api/ticket/links.rst
@@ -1,0 +1,435 @@
+Linking Tickets
+*******************
+
+Link
+====
+
+Required permission: ``ticket.agent`` **or** ``admin``
+
+Available endpoints:
+
+   GET      /api/v1/links
+   POST     /api/v1/links/add
+   DELETE   /api/v1/links/remove
+
+Request::
+
+   GET /api/v1/links
+
+   with following parameters:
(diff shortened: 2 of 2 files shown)


MrGeneration:
oh wow how to not... I messed that up quite a lot.
First of all thank you for your effort and time creating this merge request!

Long code blocks by default are provided with a "more" like feature to reduce overall page length and allow users to retrieve the information if needed:

image


The rest of the merge request did look good. 
I did correct a incorrect linking in the endpoint toctree and switched from ``::`` to ``.. code-block: json`` to allow better highlighting. You can see that in commit 2a8b8f6bf7e0d401d118a8220626a59df9928173 in case it's interesting for you. :-)

---

One last thing, as it completely doesn't fit into the rest of our documentation style.
Do you think that a endpoint listing should be available on the top of the page?

It's kinda duplicate in my opinion which is why I'd like to remove it.
image

Nitair:
Dear @MrGeneration,

thanks for your feedback 😊.

Sorry for messing up. I copied the tags article and it seems I messed up to do it correctly. I will be more careful if I do more pull requests. 

I think some sort of summary of all available endpoints is pretty useful especially if you want to search something specific without scrolling through all requests/response examples. 

It's like having a table of contents, which you wouldn't call unnecessary duplicates. 

But this is just my opinion. If you decide to not keep, you can remove that part.

MrGeneration:
Okay let's remove it for now.
If we decide that this was a bad idea then we'll add it globally to all endpoint pages.

🙏

MrGeneration:
Changes are available live already - endpoint removal will take about 5 minutes until live.

PR #208, opened 2021-11-14
Update package.rst

snc805:
For the CentOS 8 installation,the chown command will change the group to 644 which is not a valid user group. I believe it was mistakenly typed in.
The proper permission would be 755 which required by nginx to read and execute the files under /opt/zammad/public/assets folder.

### DIFF
--- a/install/package.rst
+++ b/install/package.rst
@@ -187,8 +187,7 @@ Install Zammad
 
          .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


MrGeneration (approved):

MrGeneration:
Merged - the change will be available within the next 15 to 30 minutes. 🙏

PR #210, opened 2021-12-16
Closes #209 - Debian 11 support

### LINKED ISSUE #209: Test and Verify Debian 11 package support
Workflows and required steps need verification before adding broad support etc.

MrGeneration:
This pull request adds debian 11 support and also switches to "signed-by" logics for apt like installations. 

This is due Debian 11 and Ubuntu 22 soft-deprecating `apt-key` (which throws warnings). We don't want warnings.


martini (approved):
Fine.

PR #213, opened 2022-02-23
Rework & Improve CTI API

MrGeneration:
Improving le API documentation


api/generic-cti/newcall-event.rst
```diff
@@ -0,0 +1,264 @@
...
+                  --form 'to="492214710334"' \
+                  --form 'direction="in"' \
+                  --form 'callId="25641e3f-3317-4c48-80b3-fc573c7ffe2b"' \
+                  --form 'user[]="Christopher Miller"' \
+                  --form 'user[]="Emma Taylor"'
+
+   .. tab:: URL variables
+
+      .. include:: /api/generic-cti/generic-cti_no-more-url-varibales.include.rst
+
```
YetAnotherGerrit, line 207: Add "Response" header to better outline the context switch. Maybe put the whole section before the example showcase (currently) above.

api/generic-cti/newcall-event.rst
```diff
@@ -0,0 +1,264 @@
+New Call
```
YetAnotherGerrit, line 1: New call

api/generic-cti/hangup-event.rst
```diff
@@ -0,0 +1,213 @@
+Call Hangup
```
YetAnotherGerrit, line 1: Call hangup

YetAnotherGerrit (changes requested):

api/generic-cti/newcall-event.rst
MrGeneration: I'm not sure if I can follow here?
Does this count for all 3 sub pages?

As I'm not sure what you're meaning I'm assuming you're talking about me putting the returns below the sample payloads? If so, we're doing that on the whole documentation actually:

https://docs.zammad.org/en/latest/api/user.html#create

MrGeneration:
I adjusted the rest according to your QA except for the one question I have. I hope it fits to what you meant.

api/generic-cti/newcall-event.rst
YetAnotherGerrit: Only "New call" has a specific section about the response, starting with "Depending on the chosen call direction (...)". My suggestion is to give this section a header/title "Response".

MrGeneration: I believe that "Response" as a stand alone heading without anything else is slightly unlucky when having the prior coming text in mind:

image

If you think a heading is useful at that part, how about something like `Situation specific responses` or similar?

YetAnotherGerrit: Sounds good.

MrGeneration:
Okay. From my end I'd be ready.
Did you find any objections @martini  ?

How's the result looking for you Gerrit?

YetAnotherGerrit (approved):

api/generic-cti/newcall-event.rst
```diff
@@ -0,0 +1,269 @@
...
+   * - ``callId``
+     - e.g. ``53ba82e2bd6d12d9fb2d3838f0cfb070``, ``5fb9532f40da834a``,
+       ``123456789``
+     - An ID that is unique for the call. Zammad will use this ID to identify
+       an existing call with following actions
+       (e.g. like answering or hanging up)
+
+       *This ID must be unique per call session*.
+   * - ``user``
+     - e.g. ``[John Doe]``, ``[Alice, Bob]``
```
martini, line 34: I think this should only be ``"John Doe"``, ``"Alice","Bob"``

The `[]` are only used in html form params to for the params als array.

api/generic-cti/generic-cti_no-more-url-varibales.include.rst
```diff
@@ -0,0 +1,8 @@
+🙏 Sorry, but we've decided to remove URL variable samples that were
```
martini, line 1: IMO we should remove this if we do not document this.

This way looks not good to me.

Personally (from developers view of point) I really like the old examples.

But I can go with not document it at all. But then remove it.

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
...
+     - Number that is being called
+   * - ``direction``
+     - ``in`` or ``out``
+     - The call direction - if your agent initiates a call this will be ``out``
+   * - ``callId``
+     - e.g. ``53ba82e2bd6d12d9fb2d3838f0cfb070``, ``5fb9532f40da834a``,
+       ``123456789``
+     - This is the CallID from the ``newCall`` event.
+   * - ``answeringNumber``
+     - e.g. ``42``, ``jdoe``, ``jdoe@example.com``, ``3``
```
martini, line 30: It may be also the complete phone number, depending in the BPX.

``493055571600``, ``02214710334``

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
+Call answered
```
martini, line 1: Proposal for the headline: Event: answer

api/generic-cti/hangup-event.rst
```diff
@@ -0,0 +1,215 @@
+Call hangup
```
martini, line 1: Proposal for the headline: Event: hangup

api/generic-cti/newcall-event.rst
```diff
@@ -0,0 +1,269 @@
+New call
```
martini, line 1: Proposal for the headline: Event: newCall

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
...
+   * - ``to``
+     - e.g. ``49221470334``, ``03023125771``
+     - Number that is being called
+   * - ``direction``
+     - ``in`` or ``out``
+     - The call direction - if your agent initiates a call this will be ``out``
+   * - ``callId``
+     - e.g. ``53ba82e2bd6d12d9fb2d3838f0cfb070``, ``5fb9532f40da834a``,
+       ``123456789``
+     - This is the CallID from the ``newCall`` event.
```
martini, line 28: - This is the identifier of the call session (which is also used in any other event which is related to a call session).

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
...
+.. list-table:: Available attributes and sample data for answered events
+   :widths: 20, 40, 40
+   :header-rows: 1
+
+   * - Attribute
+     - Possible value
+     - Description
+   * - ``event``
+     - ``answer``
+     - Tell Zammad there's a new call
```
martini, line 15: Tell Zammad has somebody answered this call.

api/generic-cti/hangup-event.rst
```diff
@@ -0,0 +1,215 @@
...
+.. list-table:: Available attributes and sample data for hangup events
+   :widths: 20, 40, 40
+   :header-rows: 1
+
+   * - Attribute
+     - Possible value
+     - Description
+   * - ``event``
+     - ``hangup``
+     - Tell Zammad there's a new call
```
martini, line 15: Tell Zammad somebody has hanged up the call.

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
...
+
+            .. code-block:: json
+
+               {}
+
+         Sample curl command:
+
+            .. code-block:: sh
+
+               $ curl --request POST 'https://mh-docs-dev.zammad.com/api/v1/cti/PhXp3ZG63QyJfcb1pVOnBKJWIbI' \
```
martini, line 84: {FQDN-Zammad}

api/generic-cti/answer-event.rst
```diff
@@ -0,0 +1,206 @@
...
```
martini, line 120: {FQDN-Zammad}

api/generic-cti/hangup-event.rst
```diff
@@ -0,0 +1,215 @@
...
+
+            .. code-block:: json
+
+               {}
+
+         Sample curl command:
+
+            .. code-block:: sh
+
+               $ curl --request POST 'https://mh-docs-dev.zammad.com/api/v1/cti/PhXp3ZG63QyJfcb1pVOnBKJWIbI' \
```
martini, line 98: {FQDN-Zammad}

api/generic-cti/index.rst
```diff
@@ -0,0 +1,45 @@
+Generic CTI
+***********
+
+This page describes the generic CTI API scopes and functionalities.
+
+.. include:: /api/generic-cti/cti-endpoint-limitations.include.rst
+
```
martini, line 7: If I want to find out if I want to implement this API with my BPX or if this API is the right one for me, I need to know in an introduction what I can do with this API. Technically the description if better but need to know the big picture/functional scope of the API.

This makes me miss the old introduction and feature list section here: https://docs.zammad.org/en/latest/cti/api-intro.html#introduction

MrGeneration: I personally don't like the mere introduction text with over 50% of a 2k screen about what's possible to be honest.

I also believe that we do give an idea on the situation:
image


Also, if I'd wanted to know what a specific call is capable of it should stand with the affected endpoint and not three pages before the actual relevant part is mentioned.

martini: I have no more to add from my site.

api/generic-cti/answer-event.rst
MrGeneration: What do you think, Gerrit?

api/generic-cti/hangup-event.rst
MrGeneration: What do you think, Gerrit?

api/generic-cti/newcall-event.rst
MrGeneration: What do you think, Gerrit?

api/generic-cti/answer-event.rst
MrGeneration: I'll add 2 full caller IDs, but will use german drama numbers.

api/generic-cti/newcall-event.rst
MrGeneration: It highly depends on how the call is being made.
The old documentation didn't mention it at all and when a user wanted to contribute in https://github.com/zammad/zammad-documentation/pull/159 we noticed that the wording on how to describe this situation is quite hard actually.

I currently am trying to figure out how to note that wording wise.

api/generic-cti/answer-event.rst
MrGeneration: I've taken the version from newCall because it technically describes it well enough as well in my opinion:

```
       An ID that is unique for the call. Zammad will use this ID to identify
       an existing call with following actions
       (e.g. like answering or hanging up)

       *This ID must be unique per call session*.
 ```

File: api/generic-cti/index.rst (hunk 16 above)
Comment (MrGeneration) [reply api/generic-cti/index.rst:7]: So I had a look at this topic again and this is what the condensed version looks like:
image

vs:

image

Why is it so small? Because half of the feature list describes UI functions that belong to the user documentation (and are stated there: https://user-docs.zammad.org/en/latest/extras/caller-log.html). We're trying to reduce duplicate content as much as possible.

I also dropped this line:
`Support to allow an agent to set a DND - like state` - the reason for this is that I am unable to find such a function within Zammad. While Zammad seems to have such a DND like return it currently is unclear how it can be triggered.

If this is a real function, it is currently entirely missing from both, support knowledge and documentation.

File: api/generic-cti/newcall-event.rst (hunk 4 above)
Comment (MrGeneration) [reply api/generic-cti/newcall-event.rst:34]: How about ...

```
The user(s) real name involved. You may have to provide array style (``[]``) params depending on the call method you choose.

| If the direction is ``out``, this is the name of the calling person(s).
| If the direction is ``in``, this is the name of the called person(s).

This value is optional.
```

Together with removed bracked for single values?

(instead of:)
```
The user(s) real name involved.

| If the direction is ``out``, this is the name of the calling person(s).
| If the direction is ``in``, this is the name of the called person(s).

This value is optional.
```

martini: I like our proposal a lot!

api/generic-cti/answer-event.rst
martini: In case leave it as it is.

api/generic-cti/hangup-event.rst
martini: In case leave it as it is.

api/generic-cti/index.rst
martini: I like your proposal (condensed version) here too!

api/generic-cti/answer-event.rst
YetAnotherGerrit: I prefer the "Call answered", not because it provides a better headline for the article, but it provides a better bullet point for the menu on the left.

api/generic-cti/hangup-event.rst
YetAnotherGerrit: I prefer "Call hangup", not because it provides a better headline for the article, but it provides a better bullet point for the menu on the left.

api/generic-cti/answer-event.rst
MrGeneration: Okay, we'll leave it as is then :)

api/generic-cti/hangup-event.rst
MrGeneration: Okay, we'll leave it as is then :)

MrGeneration:
@martini are you okay with the latest changes? 🙏

PR #226, opened 2022-05-04
Update for Node.js 16 dependency of Zammad 5.2.

mgruner:
- Installation instructions changed to use LTS versions of Node.js rather than the most recent.
- Removed Node.js installation instructions for package installations as Node.js gets bundled by packager.io.

### DIFF
--- a/install/includes/nodejs/centos.rst
+++ b/install/includes/nodejs/centos.rst
@@ -1,9 +1,3 @@
 .. code-block:: sh
 
-   # CentOS 7
-   $ yum install centos-release-scl
-   $ yum install rh-nodejs12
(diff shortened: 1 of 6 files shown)


mgruner:
Installation instructions changed to use LTS versions of Node.js rather than the most recent.

MrGeneration:
Looking good for me. Personally I'm wondering if we should hint that package installation already contain Node.js environments?
I believe https://docs.zammad.org/en/latest/prerequisites/software.html#node-js would be a good place?

MrGeneration:
I've added the remark for pre-bundle on package installations. If it's looking good to you, feel free to merge it into the pre-release branch.

PR #236, opened 2022-06-12
added roles endpoint

matmccabe:
Added roles endpoint, need review and some syntax troubleshooting for .rst file


MrGeneration:
Thank you for your contribution and your time.
Just to clarify. Is this a WIP PR or are you waiting for actual QA?

I've had a 10 minute look and fixed mayor issues. Already.
However, the code blocks cannot be lexed without enforcing as JSON (this is due to us providing a comment with the return code). Some of your return codes bring 4xx codes which doesn't look right. At least not if this is going to show up in the documentation. 

Some payloads send to the API and their responses look mixed up as well.
At least I can't follow them - I'm not sure what happened there. 🙏

ralf401:
Hi @matmccabe 
thanks a lot for your contribution! I redacted your draft and commited it. 

@mantas can you have a look if ok? Thank you!

api/roles.rst
```diff
@@ -0,0 +1,751 @@
...
+
+   # HTTP-Code 200 OK
+
+   {
+      "record_ids": [
+         1,
+         2,
+         3
+      ],
+      "assets": {
```
mantas, line 24: Lets make this more readable by not including `full=true` flag and `assets` content.

api/roles.rst
mantas, line 1: Minor nitpick, but other file names in this folder are singular :)

api/roles.rst
ralf401: Oh I guess it was not really correct at all in my first iteration ... Changed the request to the "standard" `/api/v1/roles` now so it should look better.

index.rst
```diff
@@ -81,8 +81,10 @@ The Zammad documentation consists of three parts:
 
    /api/intro
    /api/user
+   /api/roles
    /api/organization
    /api/group
+   /api/roles
```
mantas, line 87: I guess one of these `/api/roles` mentions is not needed?

PR #242, opened 2022-07-22
Rails console instruction for docker containers

dantefromhell:
It took me some digging through the containers to figure this out. Hopefully this helps future readers.

### DIFF
--- a/admin/console-rails-shell.include.rst
+++ b/admin/console-rails-shell.include.rst
@@ -7,4 +7,4 @@
    $ rails c
 
    # docker & docker-compose installation
-   docker run -it $NAME_OF_RAILS_CONTAINER rails r '{COMMAND}'
+   docker run -it $NAME_OF_RAILS_CONTAINER rails c
--- a/admin/console-rails-shell.include.rst
+++ b/admin/console-rails-shell.include.rst
@@ -5,3 +5,6 @@
 
    # source installation
    $ rails c
+
+   # docker & docker-compose installation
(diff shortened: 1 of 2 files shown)


MrGeneration:
Please rename your PR so that it starts with `Draft: ` until you're ready.

dantefromhell:
Apologies for the confusion, I overlooked from stuff. It's now not `Draft` status anymore.

Do you still want me to add & remove `Draft:` from the title?

admin/console-rails-shell.include.rst
```diff
@@ -5,3 +5,6 @@
 
    # source installation
    $ rails c
+
+   # docker & docker-compose installation
```
MrGeneration, line 9: Please orientate on our other console command lines. They're all starting with `$ `.
I understand that this may be irritating but is this way in the whole documentation.

Also, I'd suggest to replace the variable with a place holder like e.g. `{placeholder}`.
The variable approach may confuse our some users  as they might assume it's being set automatically.

admin/console.rst
```diff
@@ -39,6 +39,9 @@ a shell (e.g. for automation).
    # source installation
    $ rails r '{COMMAND}'
 
+   # docker & docker-compose installation
+   docker run -it $NAME_OF_RAILS_CONTAINER rails r '{COMMAND}'
```
MrGeneration, line 43: Same as above. :-)

MrGeneration (changes requested):

PR #243, opened 2022-07-29
Update contributing part

fliebe92:
We updated the contributing part to remove duplicated content (it's moved to the new developer manual in Zammad source code).


mgruner:
Thanks @fliebe92, looks good to me!

MrGeneration:
With removing `branches.rst` you're removing an argumentation option for our paid support to enforce customers to upgrade instead to ask for support for Zammad 3.5.x. We cannot remove this scope or have to move them somewhere else.

mgruner:
> With removing `branches.rst` you're removing an argumentation option for our paid support to enforce customers to upgrade instead to ask for support for Zammad 3.5.x. We cannot remove this scope or have to move them somewhere else.

@fliebe92 I would suggest to  add a new section "Supported Branches / Versions" instead where we briefly summarize the relevant info about our support scope only, and probably link to https://github.com/zammad/zammad/blob/develop/SECURITY.md.

mgruner:
@MrGeneration could you check again, please?

MrGeneration (approved):
Ready to be merged when ever you wanna.

mgruner:
🤦 oh man, I messed it up with Markdown syntax. Thanks for the fixes @MrGeneration!

MrGeneration:
No no don't be too hard to yourself. I've switches `:code:` to it's short term because it's easier to read.
The underline thing technically still works but throws ugly errors during the built process which is why I tend to fix those. :)

PR #248, opened 2022-09-30
Correct command

harrisonratcliffe:


MrGeneration:
Sorry to tell you, it's already correct.
https://docs.docker.com/compose/

PR #253, opened 2022-11-11
Introduce mysql/mariadb to pgsql-migration

YetAnotherGerrit:


appendix/migrate-to-postgresql.rst
```diff
@@ -0,0 +1,175 @@
...
+      .. code-block:: sh
+
+         $ apt update
+         $ apt install pgloader
+
+   .. tab:: CentOS
+
+      .. code-block:: sh
+
+         $ yum install -y https://download.postgresql.org/pub/repos/yum/common/redhat/rhel-7-x86_64/pgloader-3.6.2-1.rhel7.x86_64.rpm
```
MrGeneration, line 96: Something seemed to went wrong here, can you please adjust to the correct package required?
Or however you install it on CentOS.

MrGeneration (changes requested):
I haven't verified the described steps as at least 3 people at Zammad have according to your information. I thus expect them to fit ... I'm missing a lot of time resources...

What I did was to clean up *a lot* of duplicate content and inconsistencies that were also introduced during previous PRs by me.

I'm really no fan of duplicate intros.
We're repeating the same text passages on several parts of the documentation - that's useless if it's the exact same string.

Please consider using includes or choose a different wording for the migration part.
I did unify the rest, I was going to delete the whole intro text from the first paragraph but hesitated. 
image

appendix/migrate-to-postgresql.rst
YetAnotherGerrit: Works for me on CentOS7 and CentOS8 Stream on fresh Hetzner-Cloud-servers. Can you specify what went wrong?

MrGeneration (approved):

appendix/migrate-to-postgresql.rst
MrGeneration: Sorry I can't find what disturbed me yesterday, let's ignore it.

MrGeneration: Ah wait the wording thing (2 times the same text) still is valid of course.

PR #257, opened 2022-11-23
Remove MySQL/MariaDB install/config instructions.

mgruner:
We advise not to create any new MySQL installations from the 5.3 release on. This PR removes the relevant parts in the documentation, as we don't need install/configuration instructions for a DB that is not supported for new installations.


MrGeneration (changes requested):
Sorry I don't agree.
The backup script (unless you guys secretly changed it below my radar) *does* support backing up MySQL installations.

And that's good! It should do support this until Zammad 7 releases because it's *deprecated* not *not supported*. That's a difference. The backup and restore guide should state that as long as it's supported.

MrGeneration:
There are a lot of files touched even if it's just the endings. For this reason my complete review will take time until I find time for it.
Some changes looked too much for a glimpse, veto for MySQL backup instantly came.

It's a bit confusing if your client by default removes trailing spaces in lines. I do get why but it bloats up commits a lot because the documentation contains these things for age reasons.

mgruner:
@MrGeneration white space changes and backup/restore changes reverted. Now it only removes the install/configure option/instructions for MySQL on new systems.

YetAnotherGerrit (approved):

prerequisites/software.rst
```diff
@@ -163,30 +163,14 @@ installed with the Zammad-Package.
...
-   * Set ``max_allowed_packet`` to a value larger than the default of 4 MB
-     (64 MB+ recommended).
-
-   You may also want to consider the following settings for your MySQL server::
-
-      innodb_file_format = Barracuda
-      innodb_file_per_table = on
-      innodb_default_row_format = dynamic
-      innodb_large_prefix = 1
-      innodb_file_format_max = Barracuda
```
MrGeneration, line 189: Do we really want to remove a software dependency during deprecation phase?
Zammad does still allow you to run the installer that way even through we don't hint users on how.

This is kinda a bad idea and not really consistent. It also will nuke those who have to migrate systems that are not based on psql and plan on switching later. I'm fine with reducing explicit hints etc but that one's essential to me.

Or I don't get the ideas and points.

YetAnotherGerrit: After rethinking, I agree with you on this one.

mgruner: PM decided that we are not going to support new installations on MySQL. The intention of this PR is / was to make that explicit. What's the point in listing MySQL as an option for new installations here and supporting people on that? That will only lead to more migration efforts later.

MrGeneration: Right. All hints and notes on how to install this have been removed.
But above does also affect updating instances. At least from my point of view.

I'm not happy with this face right now. Like absolutely not happy.
We don't even scream at those who ignore our warnings and discouragement. But now we wanna reduce documentation.

I'd say this particular documentation part should stay as long as we're still in the grace period.
If we can't agree on that, Gerit will have to take it to PM to get a decision.

PR #259, opened 2022-12-06
Fix typo and added clarification

francoisauclair911:
Replaced `conf` with `config` as it was not the same name as described on the backup-and-restore configuration page https://docs.zammad.org/en/latest/appendix/backup-and-restore/configuration.html
That confused me as I thought maybe there was another file I was missing.
Also added the path of were this `config` file is located

### DIFF
--- a/appendix/backup-and-restore/migrate-hosts.rst
+++ b/appendix/backup-and-restore/migrate-hosts.rst
@@ -81,8 +81,8 @@ Step 6: Backup!
       :doc:`configuration page </appendix/backup-and-restore/configuration>`.
 
 Step 7: Transfer your backup files
-   You'll find the backup location within the ``conf`` file on the backup
-   directory. Make sure to adjust the backup configuration on the destination
+   You'll find the backup location within the ``config`` file on the backup
+   directory (`/opt/zammad/contrib/backup`). Make sure to adjust the backup configuration on the destination
    host according to our
    :doc:`configuration page </appendix/backup-and-restore/configuration>`
    to provide the correct backup file directory.


MrGeneration:
First of all thanks for your contribution.
I do agree on the file name typo - but am a little bit puzzled about the added directorty.

The reason for that is that the migration page technically is a red line that permanently loops you through configuration, backup and restore pages. ( https://docs.zammad.org/en/latest/appendix/backup-and-restore/configuration.html )

You actually find that reference 2 times in 2 steps close around that remark.
I'd thus suggest to not make the page longer by adding another duplicate information.
image

What do you think?

francoisauclair911:
Agreed. Would be nice if we could link the word "directory" to this section in backup configuration page. 
But that may prove a bit redundant or not very clear what we are linking as it would link the top of the page.

Screenshot 2022-12-21 at 4 54 40 PM

appendix/backup-and-restore/migrate-hosts.rst
```diff
@@ -81,8 +81,8 @@ Step 6: Backup!
       :doc:`configuration page </appendix/backup-and-restore/configuration>`.
 
 Step 7: Transfer your backup files
-   You'll find the backup location within the ``config`` file on the backup
-   directory (`/opt/zammad/contrib/backup`). Make sure to adjust the backup configuration on the destination
+   You'll find the backup location within the ``config`` file in the backup
+   directory. Make sure to adjust the backup configuration on the destination
```
MrGeneration, line 85: Sorry but revert that. You removed the default location of the most current backup version. Bad idea.

francoisauclair911: Hmm not sure I get your comment, That directory path was added by me, you then mentioned it wasn't a good idea to hardcode it. So I removed it. 
Now I should put it back ?

MrGeneration: Ugh no sorry. I've not been able to read here.

MrGeneration (approved):

MrGeneration:
Sorry for the wait. The change is life in roughly 10-30 minutes from now.

PR #260, opened 2022-12-15
Correct paths from sites-enabled to sites-available

etec-masterofsynapse:
It is bad practice to have config files directly in sites-enabled for any webserver using that directory. There should only be symbolic links there.
I changed the mentions in the docs to the sites-available directory.

### DIFF
--- a/getting-started/configure-webserver.rst
+++ b/getting-started/configure-webserver.rst
@@ -199,7 +199,7 @@ Adjusting the webserver configuration
 
          .. code-block:: sh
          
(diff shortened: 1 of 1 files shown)


MrGeneration:
Did you cross check if you found all references and the documentation logic still is given or do we have to check that?

etec-masterofsynapse:
> Thank you for your contribution. Did you cross check if you found all references and the documentation logic still is given or do we have to check that?

I checked the specific file I committed for all mentions of site-enabled and corrected them to site-available.
Since I only exchanged the word enabled to available, I don't think I messed with the docs logic.

MrGeneration (approved):
Thank you very much for your contribution

PR #263, opened 2022-12-22
Improved mysql/mariadb to postgresql migration for JSON values (zammad/zammad/issues/4431).

dominikklein:
Improved the current migration documentation with the fix for the JSON values handling (zammad/zammad/issues/4431).

There is now a rake task (when the related issue is merged) to generate the "pgloader-command" file which needs to be used so that the JSON values are migrated correctly.

### DIFF
--- a/appendix/migrate-to-postgresql.rst
+++ b/appendix/migrate-to-postgresql.rst
@@ -12,26 +12,11 @@ migration process.
    comes without any warranty. Please proceed at your own risk. In doubt
    please refer to the documentation of the tools used.
 
-.. danger:: **☠️ This path is broken for specific fields ☠️**
-
-   Please note that this migration path potentially breaks the following things
-   in Zammad 5.x:
-
-   * public links
-   * ticket / user / organization / group attributes of the type:
-
-      * multiple tree selection field
-      * multiple selection field
-
-   Please see `Issue 4431`_ for further information. If you can wait, consider
-   avoiding to migrate right now.
-
-.. _Issue 4431: https://github.com/zammad/zammad/issues/4431
 
 Preparation
 ===========
 
-#. Stop Zammad: 
+#. Stop Zammad:
 
    .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


dominikklein:
@mgruner + @MrGeneration can you do a review of the changes? When all is fine, we can merge it after the related issue from the Zammad repo was merged

MrGeneration:
Did this topic this time get a proper and intense test? I'd hate to touch this documentation part a third time.

dominikklein:
We have now a new CI-Job that tests the complete migration flow (MySQL + MariaDB) with the pgloader + local manual testing for the solution.

appendix/migrate-to-postgresql.rst
```diff
@@ -57,42 +42,9 @@ Please also have a look at :doc:`/appendix/configure-database-server`.
...
---------------------
-
-Get the credentials Zammad is currently using to access your MySQL/MariaDB
-database from ``config/database.yml``. You will need them later.
-
-Now adjust the configuration file to fill in the credentials for your new
-PostgreSQL server. Use ``postgresql`` as ``adapter``.
 
-
-.. include:: /install/includes/postgres-permissions.rst
```
MrGeneration, line 70: This is actually a very important hint for the `rake db:create` part.
Removing this may leave people "not knowing what to do or why it fails" depending on the systems configuration.

As this part here expects the user to install postgresql and doesn't tell anything on "how to create the zammad db user" (which is fine because we expect the admin to know), the process will most likely fail. The default behavior of postgresql is to not allow database creation for normal users unless stated other wise.

MrGeneration:
I need much patience and time for this. it's hard to track the changes and my mind doesn't let me right now.
ETA 7 working days.

appendix/migrate-to-postgresql.rst
```diff
@@ -117,70 +69,105 @@ Install pgloader and migrate
...
-		INTO pgsql://zammad:pgsql_password@localhost/zammad
+      .. code-block:: sh
 
-   ALTER SCHEMA 'zammad' RENAME TO 'public'
+         $ su - zammad
+         $ rake zammad:db:pgloader > /tmp/pgloader-command
 
-   WITH BATCH CONCURRENCY = 1;
 
+Afterwards you need to align the created file with the needed information.
```
mgruner, line 93: @dvuckovic it might be good to show an example here for what needs to be changed. I guess credentials have to be changed almost always for production instances.

mgruner (approved):
Looks good to me in general, please see one suggestion to aid the user in preparing the command file.

appendix/migrate-to-postgresql.rst
dvuckovic: There is already a helpful comment in the command file just above the line with the target PostgreSQL URL which has to be tweaked:

```
LOAD DATABASE
  FROM mysql://root:zammad@127.0.0.1/zammad_development

  -- Adjust the PostgreSQL URL below to correct value before executing this command file.
  INTO pgsql://zammad:zammad@localhost/zammad_development

ALTER SCHEMA 'zammad_development' RENAME TO 'public'
...
```

The rest of the values in the file should be current, they are automatically read from the database configuration of the current environment.

I will add the same info in the PR.

MrGeneration (approved):
I had to correct some intend and spacing errors but now it looks alright to me.

Please note:
I did not verify the convert process itself, if I have to verify that as well let me know. I only verified the described steps to get to the point of conversion.

PR #266, opened 2023-01-12
Maintenance: Change to the FTP download URL now that we have continuous delivery for source archives

mgruner:
`zammad-latest.*` on FTP will be updated by CI whenever a commit is applied to `stable`. Let's update the docs for this as well.

### DIFF
--- a/install/source/include-get-the-source.rst
+++ b/install/source/include-get-the-source.rst
@@ -1,5 +1,5 @@
-Get the latest stable release of Zammad `here <https://github.com/zammad/zammad/archive/stable.tar.gz>`_,
-or find the initial version at https://ftp.zammad.com.
+Get the latest stable release of Zammad `here <https://ftp.zammad.com/zammad-latest.tar.gz>`_.
+This file will be updated whenever new bug-fixes are applied, so you can update from this URL regularly.
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


install/source/include-get-the-source.rst
```diff
@@ -1,5 +1,5 @@
-Get the latest stable release of Zammad `here <https://github.com/zammad/zammad/archive/stable.tar.gz>`_,
-or find the initial version at https://ftp.zammad.com.
+Get the latest stable release of Zammad `here <https://ftp.zammad.com/zammad-latest.tar.gz>`_.
+This file will be updated whenever new bug-fixes are applied, so you can update from this URL regularly.
 
 .. code-block:: sh
```
MrGeneration, line 5: Is it intended that we leave the Github URL below as is and remove the Github part? 
I mean technically both works, I'm not sure if that's the best approach.

We could actually hint for both approaches (e.g. in case `ftp.zammad.com` is not available for a moment) and could also decide what the "default" as per documentation to use should be. I don't have a real preference but right now this feels odd.

MrGeneration (changes requested):

install/source/include-get-the-source.rst
mgruner: Good catch @MrGeneration. That was not intended, I updated the snippet to use the FTP URL as well.

We should only use that one as the GitHub download has different content (e.g. cleanup.sh is not executed).

Can you check again, please?

PR #270, opened 2023-01-15
Add part about cascaded queries and .keyword modifier

bogatkaj:
After searching the web for hours I fond those two options to modify the search queries the way I need them.
To prevent others from having this pain of search I added the to the documentation.

### DIFF
--- a/api/intro.rst
+++ b/api/intro.rst
@@ -619,6 +619,22 @@ The value of this header can be one of the following:
 
 On behalf of is available for all endpoints.
 
+Advanced search queries
+=======================
+
+For searching for multiple parameter:value pairs with restriction to match all parameters, they can be AND-cascaded by using ``%20AND%20``
+
+So a query would look like this:
+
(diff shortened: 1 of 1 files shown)


MrGeneration:
First of all, thank you for your contribution!

Personally I'm not a big fan of duplicate information - this PR mainly adds this kind of information to the doc.
It mostly contains content which is covered in the user documentation:
https://user-docs.zammad.org/en/latest/advanced/search.html

I'd personally prefer to extend faulty pages (in case that's the case in your opinion) and link it on other places needed.
That would allow you to also reference, fairly short, on the include that's shown on every search section:
https://github.com/zammad/zammad-documentation/blob/main/api/includes/sort_and_order.rst

You could even use a totally new include for that.

The intro is fairly chunky already and I tried to remove information from it that were relevant on very specific parts only whenever possible and instead mention it on the part it's relevant.

What's your opinion on that approach?

PR #271, opened 2023-01-19
Add a note regarding newer imlib2

mantas:
This adds a note to documentation regarding photo orientation issue fixed in this MR:

https://git.zammad.com/zammad/zammad/-/merge_requests/4616

### DIFF
--- a/prerequisites/software.rst
+++ b/prerequisites/software.rst
@@ -155,6 +155,12 @@ installed with the Zammad-Package.
 
 .. note::
 
+   | Currently LTS Linux distributions include an outdated version of imlib2.
+   | If you installed imlib2 1.9 or newer from other sources, please set ``ZAMMAD_RSZR_NO_AUTOROTATE`` environment variable to true.
+     Otherwise there may be issues with orientation of photos.
+
+.. note::
+
    | ``libimlib2-dev`` **or** ``imlib2-devel`` are no longer required.
    | *However:* If you have to use ``bundle install`` for e.g. custom Gems or
      development, you'll need to install it!


MrGeneration:
I think this kind of a warning is a perfect warning for the postinstall script in fancy red.
From what I've seen and learned in my past 4.5 years, people don't read the software page - especially not *after* initial installation.

MrGeneration:
> As for postinstall, should I append it to [final_message](https://github.com/zammad/zammad/blob/4ed6c526f7cb7f56b9cdc07182ba4752f85651e6/contrib/packager.io/functions#L325)?

Yes combined with a check for that environment possibly. We do that for e.g. Elasticsearch already:
As for postinstall, should I append it to [final_message](https://github.com/zammad/zammad/blob/4ed6c526f7cb7f56b9cdc07182ba4752f85651e6/contrib/packager.io/functions#L325)?

Yes you're right, this sadly doesn't affect or help source, docker and helm users.
Especially with docker it doesn't have any affect because the container has to ship it.

Let's leave it as is in the doc, I'll QA it later and check if we should add it to more places.

mantas:
> Yes combined with a check for that environment possibly

I remember discussed detecting imlib2 version. But then decided against it since `pkg-config` may be not available on production systems. I guess we got same issue here. I'll add a static text then in the main MR.

MrGeneration:
> I remember discussed detecting imlib2 version. But then decided against it since pkg-config may be not available on production systems. I guess we got same issue here.

I would have gone with checking the ENV for being absent and then showing that message like "check if you have v2 and if not consider setting this variable"

But if that's too much it's alright, it's just an idea.

mantas:
This is a very cool idea if we'll end up in a situation that some distros offer the newer version while others don't.

But for now it seems to be overkill to have package-post-installation message when all distros we support still use an outdated imlib2.

mantas:
We found a different way to approach this issue and such documentation change is no longer needed. Closing this PR without merging.

PR #272, opened 2023-01-24
Include a hint that Redis will be a regulard dependency from Zammad 6.0 onwards.

mgruner:
There will also be announcements on other channels, but the documentation should also give a hint about it.

This can be merged after review, don't need to wait for Zammad 5.4.

Not sure if we can/need to make this more prominent on the software requirements page somehow.

### DIFF
--- a/prerequisites/software.rst
+++ b/prerequisites/software.rst
@@ -289,11 +289,6 @@ An Elasticsearch plugin is required to index the contents of email attachments:
 
    We consider this topic as :ref:`performance_tuning`.
 
-.. hint:: **Redis will become a regular dependency in Zammad 6**
-
-   Starting with Zammad 6.0, Redis will be required to operate Zammad,
-   not just as a performance optimization.
-
 2.8.1 Redis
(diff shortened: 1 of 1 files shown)


prerequisites/software.rst
```diff
@@ -289,6 +289,10 @@ An Elasticsearch plugin is required to index the contents of email attachments:
 
    We consider this topic as :ref:`performance_tuning`.
 
+.. hint:: **Redis will become a regular dependency in Zammad 6**
```
MrGeneration, line 292: Generally looking good to me.

Just a suggestion:
I'd suggest to move the whole hint block from line `292` below line `307` with the same indent.
I think it makes it much more visible that it doesn't affect memcache (even though we mention it)

image

vs

image

MrGeneration (approved):

prerequisites/software.rst
mgruner: Feel free to change it this way and merge it, please!

PR #276, opened 2023-02-20
Add documentation for ui_ticket_add_article_hint

dvuckovic:
Provides an addition to the documentation explaining the usage of the newly added [`ui_ticket_add_article_hint`](https://github.com/zammad/zammad/pull/4326) hidden setting.


MrGeneration:
Looking good to me. The only exception is the image, I'll change it to a gif showing all scenarios shortly.

MrGeneration (approved):
Sorry I did find something else.
I switched `outgoing call` to `public call`. I believe outgoing is slightly miss leading because Zammad doesn't really differ on that regard in that context. UI wise the only thing that's different in that scope is if the customer can see the article or not.

I also added email scope because I have the slight feeling that people will ask other wise...

Please double tab, if everything looks good from your point of view hit the merge button. 🙏

PR #277, opened 2023-02-28
Add SLES15 support

MrGeneration:


install/includes/postgres-installation.rst
```diff
@@ -32,5 +32,7 @@
 
          $ zypper refresh
          $ zypper install postgresql postgresql-server postgresql-contrib
+         # openSuSE 15 also requires:
```
mgruner, line 35: Should this be SLES instead?

PR #278, opened 2023-03-09
Change docker compose env variables to table format

sangramrath:
Change docker compose env variables into a table format for easy readability.
Attaching a screenshot for visual reference
Screenshot 2023-03-09 at 6 17 34 PM

### DIFF
--- a/appendix/configure-env-vars.rst
+++ b/appendix/configure-env-vars.rst
@@ -75,7 +75,7 @@ RAILS_LOG_TO_STDOUT
 ZAMMAD_BIND_IP
    The IP address that the web server is bound to.
 
-   Default: ``0.0.0.0``
+   Default: ``127.0.0.1``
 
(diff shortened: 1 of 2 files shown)


MrGeneration:
I must confess that approach as it is right now looks much worse in the table.

Possibly three columns would be better.

sangramrath:
I thought about it, but that would mean the tips / info would have to be along side the description (below it mostly). Here's a screenshot, looks good?
Screenshot 2023-03-09 at 7 23 39 PM

ralf401:
Hi @sangramrath 

thanks for your contribution! I think your approach is more readable as the current state. I prefer the 4 column layout, of which you created your PR. I just prefer to have the same columns of the different tables with the same width. I merge it and do some small adjustments then.

ralf401 (approved):

PR #279, opened 2023-03-10
Fixes invalid endpoint documentation

frosenek:
The route to delete an organization is ``/api/v1/organizations/{id}`` instead of ``/api/v1/organization/{id}``


MrGeneration (approved):
Please always direct your PR to pre-release.

PR #286, opened 2023-03-24
Add websockets port to software requirements docs

mzur:
I was missing the port that is used for websockets connections on this documentation page. Is `6042` the correct port? I found it on [adjusting the webserver configuration](https://docs.zammad.org/en/latest/getting-started/configure-webserver.html?highlight=websockets#adjusting-the-webserver-configuration) (local testing or other proxy servers).

### DIFF
--- a/prerequisites/software.rst
+++ b/prerequisites/software.rst
@@ -45,7 +45,7 @@ requirements for your clients. This ensures that Zammad works as expected.
 1.2. Network requirements
 -------------------------
 
-Zammad uses web sockets. Some application firewalls may filter these
+Zammad uses web sockets (listening on port ``6042``). Some application firewalls may filter these
 connections. This may lead to decreased browser performance.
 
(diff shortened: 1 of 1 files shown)


MrGeneration:
Thanks for your contribution.
The thought is generally good, however, a bad idea because you're using a hard coded information for something that isn't hard coded. See: https://docs.zammad.org/en/latest/appendix/configure-env-vars.html#network-options

Also, the noted port is technically incorrect because that's the local level of the machine and not what falls out of your web server.

mzur:
So which port should I allow in my firewall then (assuming a stock install)? That's the information I'm missing in the docs.

MrGeneration:
Our documentation is not made / there to teach you basics.
Open wikipedia and have a look how websockets work, you may understand why our documentation hints it but doesn't mention port numbers.

It mainly affects application aware firewalls. For further input, *please* go to the community.

PR #287, opened 2023-04-18
Document redis as a regular dependency.

mgruner:


mgruner:
@MrGeneration do you think this would be sufficient? Package installations will handle this automatically without user interaction.

mgruner:
@dvuckovic FYI

MrGeneration:
We may want to add the installation steps - I can handle that. Shouldn't be too much.
Is a specific minimum or maximum redis version required or does it not matter?

> Package installations will handle this automatically without user interaction.

You mean as package dependency?
Is this change already active on develop for verification reasons?

mgruner:
> We may want to add the installation steps - I can handle that. Shouldn't be too much. Is a specific minimum or maximum redis version required or does it not matter?

There is no minimum version currently. Which installation steps do you mean, to install redis on source installation systems? Should we not leave that to the user?

To configure Zammad there are no steps needed. It will assume a `REDIS_URL=redis://localhost` if none is set for action cable (but not for the legacy WS server, this will continue to use the file back end unless `REDIS_URL` is explicitly set).

> > Package installations will handle this automatically without user interaction.
> 
> You mean as package dependency? Is this change already active on develop for verification reasons?

Correct. I tested it with installation and update from previous `stable`, and it worked well for me. No change in installation procedure neccessary.

MrGeneration:
> Correct. I tested it with installation and update from previous stable, and it worked well for me. No change in installation procedure neccessary.

On *all* packages?
If not I will.

> To configure Zammad there are no steps needed. It will assume a REDIS_URL=redis://localhost if none is set for action cable (but not for the legacy WS server, this will continue to use the file back end unless REDIS_URL is explicitly set).

That sounds like a crucial step that should appear in the required steps to make it run or set it via `postinstall.sh` if not set like we do with STDOUT logging. Configuration of redis is out of scope, ensuring you can solve dependencies is not.

mgruner:
> > Correct. I tested it with installation and update from previous stable, and it worked well for me. No change in installation procedure neccessary.
> 
> On _all_ packages? If not I will.

No, I tested on Ubuntu 22.04 LTS to make sure it works in general with the new redis dependency.

BTW relevant change was in https://github.com/zammad/zammad/commit/52306c6e038f5336b90b0974b0ae497a6d63bafc

> > To configure Zammad there are no steps needed. It will assume a REDIS_URL=redis://localhost if none is set for action cable (but not for the legacy WS server, this will continue to use the file back end unless REDIS_URL is explicitly set).
> 
> That sounds like a crucial step that should appear in the required steps to make it run or set it via `postinstall.sh` if not set like we do with STDOUT logging. Configuration of redis is out of scope, ensuring you can solve dependencies is not. So it would at least belong into the source code installation part to give users enough guidence. Copy cats won't succeed other wise.

Feel free to add what you see fit, and I can review it.

My point was that for single-machine package based installations (which should be the majority), no changes are needed and `REDIS_URL` does not need to be set. Thus no relevant changes in the .pkgr.yml or related files. Martin wanted the legacy WS server to keep its behaviour in this scenario (using files).

dvuckovic (approved):

PR #292, opened 2023-04-21
Fixes #290 - Add Zammad safe mode.

### LINKED ISSUE #290: Add Zammad safe mode
Recently, a feature was introduced named Zammad safe mode. This allows running any rails app except rails server without requiring any third-party service like Redis. This must be documented within the Zammad documentation.

dvuckovic:
Changelog:

* Add a warning block explaining `ZAMMAD_SAFE_MODE` environment variable with sample output.

### DIFF
--- a/admin/console.rst
+++ b/admin/console.rst
@@ -58,17 +58,16 @@ This reduces loading times greatly.
    You may receive errors and console will refuse to start in case they are not
    available.
 
-   However, it's possible to start rails console in safe mode by setting a
-   special environment variable. With ``ZAMMAD_SAFE_MODE=1`` set, availability
-   of these services will be ignored.
+   However, it's possible to start rails console in safe mode by setting
+   :ref:`a special environment variable <safe_mode>`. With
+   ``ZAMMAD_SAFE_MODE=1`` set, availability of these services will be ignored.
 
    .. tabs::
 
       .. code-tab:: console Package Installation
 
          $ ZAMMAD_SAFE_MODE=1 zammad run rails c
-
-         Zammad app is running in safe mode. Any third-party service is ignored.
+         Zammad is running in safe mode. Any third-party services like Redis are ignored.
 
          There was an error trying to connect to Redis via redis://localhost:6379.
          Please provide a Redis instance at localhost:6379 or set REDIS_URL to point to a different location.
@@ -79,8 +78,7 @@ This reduces loading times greatly.
       .. code-tab:: console Source Installation
 
          $ ZAMMAD_SAFE_MODE=1 rails c
-
-         Zammad app is running in safe mode. Any third-party service is ignored.
+         Zammad is running in safe mode. Any third-party services like Redis are ignored.
 
          There was an error trying to connect to Redis via redis://localhost:6379.
          Please provide a Redis instance at localhost:6379 or set REDIS_URL to point to a different location.
--- a/admin/console.rst
+++ b/admin/console.rst
@@ -51,6 +51,29 @@ This reduces loading times greatly.
 
 .. include:: /admin/console-rails-shell.include.rst
 
+.. warning:: **Safe Mode**
+
+   Normally, starting rails console requires certain `third-party services`_ to
+   be up and running. You may receive errors and console will refuse to start
+   in case they are not available.
+
+   However, it's possible to start rails console in safe mode by setting a
+   special environment variable. With ``ZAMMAD_SAFE_MODE=1`` set, availability
+   of these services will be ignored:
+
+   .. code-block:: console
+
+      $ ZAMMAD_SAFE_MODE=1 rails c
+      Zammad app is running in safe mode. Any third-party service is ignored.
+
+      There was an error trying to connect to Redis via redis://localhost:6379.
+      Please provide a Redis instance at localhost:6379 or set REDIS_URL to point to a different location.
+      ＃<Redis::CannotConnectError: Error connecting to Redis on localhost:6379 (Errno::ECONNREFUSED)>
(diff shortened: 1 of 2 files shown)


admin/console.rst
```diff
@@ -51,6 +51,29 @@ This reduces loading times greatly.
...
+      $ ZAMMAD_SAFE_MODE=1 rails c
+      Zammad app is running in safe mode. Any third-party service is ignored.
+
+      There was an error trying to connect to Redis via redis://localhost:6379.
+      Please provide a Redis instance at localhost:6379 or set REDIS_URL to point to a different location.
+      ＃<Redis::CannotConnectError: Error connecting to Redis on localhost:6379 (Errno::ECONNREFUSED)>
+      Loading production environment (Rails 6.1.7.3)
+      3.1.3 :001 >
+
+.. _`third-party services`: Server requirements
```
MrGeneration, line 75: That's a dead link. What did you want to link exactly?

admin/console.rst
```diff
@@ -51,6 +51,29 @@ This reduces loading times greatly.
 
 .. include:: /admin/console-rails-shell.include.rst
 
+.. warning:: **Safe Mode**
```
MrGeneration, line 54: I think ``warning`` is a bit agressive for a hint / note for people not solving their dependencies.
Maybe a ``hint`` would be an fitting alternative

admin/console.rst
```diff
@@ -51,6 +51,29 @@ This reduces loading times greatly.
...
+   be up and running. You may receive errors and console will refuse to start
+   in case they are not available.
+
+   However, it's possible to start rails console in safe mode by setting a
+   special environment variable. With ``ZAMMAD_SAFE_MODE=1`` set, availability
+   of these services will be ignored:
+
+   .. code-block:: console
+
+      $ ZAMMAD_SAFE_MODE=1 rails c
```
MrGeneration, line 66: This is a very fixed example for source code and docker variants only. Maybe wildcarding it or something might be a good idea. Or at least note that it depends on the used system.

People tend to be *very* confused on what variant to use especially if you don't push their head into the differences.

admin/console.rst
dvuckovic: Fixed.

Links to Software requirements page, where different services are listed.

admin/console.rst
dvuckovic: Fixed.

Switched to using tabs with separate code blocks.

tschaefer:
Actually, the `ZAMMAD_SAFE_MODE` supports any Rails app besides `rails server`. Maybe this should be reflected in the documentation, but we stuck with the `rails console` examples.

MrGeneration:
> Actually, the ZAMMAD_SAFE_MODE supports any Rails app besides rails server. Maybe this should be reflected in the documentation, but we stuck with the rails console examples.

imho a good point. I personally expected it with the environmental part:
https://docs.zammad.org/en/latest/appendix/configure-env-vars.html

But then I also saw dusan mentioning rails console only and thought it might be good if users wouldn't run productive systems that way. At least the way I understand that it technically would potentially break Zammad operation...

dvuckovic:
@tschaefer 

> Actually, the `ZAMMAD_SAFE_MODE` supports any Rails app besides `rails server`. Maybe this should be reflected in the documentation, but we stuck with the `rails console` examples.

Currently, docs do not define a term called "Rails app", so I doubt the admins would know what this relates to. As far as they are concerned, there is only one app: Zammad.

Rather than documenting theoretical capabilities, I felt like the best course of action would be to document real-life use cases of this new ENV variable. That said, if you have some more for me, I'd be happy to add them to appropriate places. I could only positevely identify one so far (I don't think generation of the translation catalog relates to normal operation).

@MrGeneration 

> imho a good point. I personally expected it with the environmental part: https://docs.zammad.org/en/latest/appendix/configure-env-vars.html
> 
> But then I also saw dusan mentioning rails console only and thought it might be good if users wouldn't run productive systems that way. At least the way I understand that it technically would potentially break Zammad operation...

This is a good point as well. I can add it in that section as well, unless you think it's not a good idea due to your concerns. Just let me know.

MrGeneration:
Sorry Dusan I didn't think I have power on that regard.

> This is a good point as well. I can add it in that section as well, unless you think it's not a good idea due to your concerns. Just let me know.

I personally am not a fan of duplicate information. For me it belongs to the environment variable section (because technically it is). One could hint with temporary environment variables in the console section. But that hint should be like a one line and possibly more of a "wildcard" variant.

I think deciding that is out of my responsibilities right now.

dvuckovic:
@MrGeneration Thanks for the feedback, I appreciate it!

Personally, I don't think it's a good idea to mention it just under environment variables. It might be a common mistake for admins to try to start the console without all required services in place, so it's a good idea to bring attention to this right where we are describing the usage of the console.

I was thinking only to mention it in some additional words in the environment variables. There we could explain that it covers many "rails apps", not just the console. But I still don't like the term, we should find something more suitable and self-explanatory.

MrGeneration:
Again, just my opinion: The mayority of information of this thingie belongs to environment. I'm not against noting it on console section at all as long as it's not duplicate content.

> There we could explain that it covers many "rails apps", not just the console.

We're documented Zammad. I don't think extending the "visual" scope to other rails apps and thus I'd use "Zammad" as synonym for "rails app". But that's just me.

dvuckovic:
@tschaefer Ping!

dvuckovic:
@mgruner Could you please take a look at this PR instead of Tobias? Thanks!

mgruner (approved):
Looks very nice!

PR #298, opened 2023-06-14
Add base URL tip in DB init section.

tschaefer:
See commit https://github.com/zammad/zammad/commit/30ea2c309f2bbf89d38dbd54a5e88d110f4bca7b

### DIFF
--- a/install/includes/db-seed-base_url.rst
+++ b/install/includes/db-seed-base_url.rst
@@ -1,6 +1,0 @@
-.. tip::
-
-   **🤓 Preserve a restart ...**
-
-   You can set the base URL of your Zammad installation by setting the
-   ``ZAMMAD_HTTP_TYPE`` and ``ZAMMAD_FQDN`` environment variables.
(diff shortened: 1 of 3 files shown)


install/includes/db-seed-base_url.rst
```diff
@@ -0,0 +1,6 @@
+.. tip::
+
+   **🤓 Preserve a restart ...**
+
+   You can set the base URL of your Zammad installation by setting the
+   ``ZAMMAD_HTTP_TYPE`` and ``ZAMMAD_FQDN`` environment variables.
```
dvuckovic, line 6: Without an example, I'm not sure how would an admin know what to set these variables to?

Perhaps we can document them like others in `appendix/configure-env-vars.rst` and then link from here?

install/source.rst
```diff
@@ -227,6 +227,8 @@ Step 4: Initialize your database
 
    Ensure to do this as ``zammad`` user in your Zammad directory!
 
+.. include:: /install/includes/db-seed-base_url.rst
+
```
dvuckovic, line 231: If there is only a single inclusion of this file, I'd like to argue we don't really need this as a separate file. Modularity is not really the goal here, we leverage `include` to avoid repetition only.

dvuckovic (changes requested):

PR #301, opened 2023-06-30
Added hint for Debian 12.

mgruner:
This should go straight to `main`, as it is already available.

### DIFF
--- a/install/package.rst
+++ b/install/package.rst
@@ -60,29 +60,29 @@ Add Repository
                $ curl -fsSL https://dl.packager.io/srv/zammad/zammad/key | \
                  gpg --dearmor | tee /etc/apt/trusted.gpg.d/pkgr-zammad.gpg> /dev/null
 
-            Ubuntu 16.04
-               .. code-block:: sh
+         Ubuntu 16.04
+            .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


install/package.rst
```diff
@@ -104,6 +104,12 @@ Add Repository
                   $ echo "deb [signed-by=/etc/apt/trusted.gpg.d/pkgr-zammad.gpg] https://dl.packager.io/srv/deb/zammad/zammad/stable/debian 11 main"| \
                     tee /etc/apt/sources.list.d/zammad.list > /dev/null
 
+            Debian 12
+               .. code-block:: sh
+
+                  $ echo "deb [signed-by=/etc/apt/trusted.gpg.d/pkgr-zammad.gpg] https://dl.packager.io/srv/deb/zammad/zammad/stable/debian 12 main"| \
+                    tee /etc/apt/sources.list.d/zammad.list > /dev/null
+
```
dvuckovic, line 112: This looks fine, for sure, but I think all these blocks have unnecessary indentation. They are rendered under **Install Repository Key** subtitle, but they should be on the same level.

I'm fine with postponing this change for later, it just looks strange to me.

dvuckovic (approved):

install/package.rst
mgruner: I'll check and update, sorry about that.

PR #314, opened 2023-07-26
Maintenance: Add CI pipeline based on Github workflows.

dvuckovic:
Job example (from the fork): [pre-release](https://github.com/dvuckovic/zammad-documentation/actions/runs/5666836895)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,20 +28,15 @@ on:
         type: string
 
 jobs:
-  test:
-    name: Test
-    runs-on: ubuntu-20.04
+  build:
+    name: Build
+    runs-on: ubuntu-latest
     steps:
     - name: Checkout Repository
       uses: actions/checkout@v3
       with:
         fetch-depth: 50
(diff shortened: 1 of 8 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,100 @@
...
+        type: string
+      requirements_path:
+        description: Path to the `requirements.txt`, relative to the repository root.
+        required: false
+        default: requirements.txt
+        type: string
+
+jobs:
+  build:
+    name: Build
```
mgruner, line 32: Maybe `Lint & Build` or `Test & Build` or similar?

mgruner (approved):
Looks amazing, and I guess we can improve it as we go.

.github/workflows/docs.yml
dvuckovic: I tried all of those, Github will ellipsize it anyway in most views. Better to be concise in this case and we can improve it later if we have additional jobs.

PR #315, opened 2023-07-26
Maintenance: Add CI pipeline based on Github workflows.

dvuckovic:
Job example (from the fork): [main](https://github.com/dvuckovic/zammad-documentation/actions/runs/5666763737)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,100 @@
+name: Documentation
+
+# Controls when the workflow will run.
+on:
+
+  # Allow the workflow to be reusable.
+  workflow_call:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,100 @@
...
+        type: string
+      requirements_path:
+        description: Path to the `requirements.txt`, relative to the repository root.
+        required: false
+        default: requirements.txt
+        type: string
+
+jobs:
+  build:
+    name: Build
```
mgruner, line 32: Consider renaming as well

mgruner (approved):

PR #328, opened 2023-08-16
Fixes #323 - Redacted database installation in Zammad source installa…

### LINKED ISSUE #323: Install from source: database installation
At the very beginning of the "from source" instructions it says "ensure to provide your database server and web server at this point".
Then it says "Below commands do neither include the database server nor the web server".

And then the first thing that comes is "apt install postgresql postgresql-contrib".

This does not fit. We should either include the database installation part and correct the two introductions that currently say otherwise or not include the database installation.

ralf401:
…tion


mgruner (approved):
Looks nice!

PR #329, opened 2023-08-17
Add warning about upcoming linux distribution support change.

mgruner:


mgruner:
@ralf401 can you check this, please? I found that Ubuntu 16.04 was listed, even though it is already not supported for 6.0, so I removed it. I guess we don't need to also do that in `main`, otherwise there might be conflicts. Nobody should use 16.04 for new installations, I guess.

ralf401 (approved):
Looks good. I just noticed an already existing typo in the linked appendix/includes/mysql-deprication-note.rst (dopped -> dropped), but I can fix this as well after merge.

PR #331, opened 2023-08-18
Fixed typo / changed latest stable to 6.0

ralf401:


MrGeneration:
Seems like you got confused here too?
The 5.1 part is an example only. I wanted to have an example for people to understand the path but didn't want to keep updating this exact  phrase on every release...

install/update.rst
```diff
@@ -19,11 +19,11 @@ Updating Zammad
       for your and your customers safety.
 
       In case you couldn't update for a longer time, please ensure to at least
-      update from **mayor to mayor** version. Big version jumps *may* work but
+      update from **major to major** version. Big version jumps *may* work but
       usually go terribly wrong. As example, expecting the current stable
       version of Zammad being 5.1 and your instance being on Zammad 2.4, your
       path would look like so:
-      ``2.4`` → ``3.0`` → ``4.0`` → ``5.0`` → ``latest stable (5.1)``
+      ``2.4`` → ``3.0`` → ``4.0`` → ``5.0`` → ``latest stable (6.0)``
```
mgruner, line 26: This is not correct now, the text says 'being 5.1'. Maybe you can just keep this section as before, or update both places.

install/update.rst
```diff
@@ -19,11 +19,11 @@ Updating Zammad
       for your and your customers safety.
```
mgruner, line 19: `for customers'`

mgruner (approved):
Nice fixes, just minor suggestions.

PR #334, opened 2023-08-21
Fixes #305 - Added: Shared Draft API

### LINKED ISSUE #305: Shared draft API is missing

ralf401:


api/ticket/shared_drafts.rst
```diff
@@ -0,0 +1,1838 @@
...
+
+   {
+      "shared_draft_id": 2,
+      "assets": {
+         "TicketSharedDraftZoom": {
+            "2": {
+               "id": 2,
+               "ticket_id": 58,
+               "new_article": {
+                  "body": "<div>Some text for a shared draft. <br></div><div><br></div><div>Some inline image:<img tabindex=\"0\" style=\"width: 1000px; max-width: 100%;\" src=\"data:image/jpeg;base64,/9j/4AAQSkZJRg[...]
```
mgruner, line 26: Cutting off withouth the final `",` confuses the syntax highlighting in the final output. Please correct this (`[...]",`)

api/ticket/shared_drafts.rst
```diff
@@ -0,0 +1,1838 @@
+Shared drafts
```
mgruner, line 1: Title Case: `Shared Drafts`?

api/ticket/shared_drafts.rst
```diff
@@ -0,0 +1,1838 @@
...
+*************
+
+Show
```
mgruner, line 4: There is one additional currently undocumented route, `import_attachments`. Maybe you can check with @mantas if we should document this as well, and how it works.

mgruner (approved):
Looks nice, minor suggestions to consider.

api/ticket/shared_drafts.rst
ralf401: Some day I will learn it ;-)

api/ticket/shared_drafts.rst
ralf401: I got in touch with him and we considered it as not important. He states that there is probably just one use case which is not very feasible (user opens ticket and draft > copies the form ID > calls the API with this ID to import an attachment).

But thanks for your review and keeping your eyes open anyway :-)

PR #337, opened 2023-08-22
Fixes #262 - Redacted ticket and user deletion

### LINKED ISSUE #262: Privacy appendix outdated
Privacy appendix still mentions "no user deletion via admin panel" which no longer is valid.

image

ralf401:
Due to the already available [documentation in the admin part](https://admin-docs.zammad.org/en/latest/system/data-privacy.html), I kept it short and linked to that part.

### DIFF
--- a/appendix/configure-database-server.rst
+++ b/appendix/configure-database-server.rst
@@ -1,24 +1,24 @@
 Configure Database server
 *************************
 
-.. note:: 
+.. note::
 
-   Parts of this page also applies to both supported database servers. 
+   Parts of this page also applies to both supported database servers.
    We can't provide a complete how to and will only enlighten the relevant parts for Zammad.
 
-Within ``database.yml`` (``config/`` directory) you can define the allowed pool size. 
+Within ``database.yml`` (``config/`` directory) you can define the allowed pool size.
 By default each Zammad process takes up to ``50`` connections (``pool: 50``).
 
-This should be fairly enough for *every* use case. 
-If you experience database connection timeouts or similar pool errors, this usually 
+This should be fairly enough for *every* use case.
+If you experience database connection timeouts or similar pool errors, this usually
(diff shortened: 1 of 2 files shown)


appendix/configure-database-server.rst
```diff
@@ -27,7 +27,7 @@ Keep in mind that versions may differ from your setup - adapt where needed.
 
       .. code-block::
 
-         /etc/postgresql/(10|11|12)/main/postgresql.conf
+         /etc/postgresql/(10|11|12|13|14|15)/main/postgresql.conf
```
mgruner, line 30: Looks good, but we will have to keep updating this one every few months. What about `/*/` or similar instead?

appendix/privacy.rst
```diff
@@ -11,14 +11,15 @@ The following kinds of data are stored locally on the production system:
...
-   To manually delete users and all their associated tickets
-   (*e.g.,* in compliance with a “Right to Forget” request under the GDPR),
-   :doc:`use the console </admin/console/dangerzone-for-experts>`.
 
-   .. note:: The ability to delete users via the admin panel
-      is planned for a future release of Zammad.
+   To **manually** delete users and all their associated tickets
+   (*e.g.* in compliance with a “Right to Forget” request under the GDPR),
+   you can use the
+   `functions <https://admin-docs.zammad.org/en/latest/system/data-privacy.html>`_
```
mgruner, line 20: To be more precise, perhaps `data privacy functions` rather than just `functions`?

mgruner (approved):
Nice! Please see small suggestions.

appendix/configure-database-server.rst
ralf401: You are right. I suggest something like /etc/postgresql/{your version}/main/postgresql.conf

PR #340, opened 2023-08-24
Document 'How to run rails/rake commands in containers'.

mgruner:
Follow-up for https://github.com/zammad/zammad-docker-compose/pull/366.

This should go to main, because it is already valid for the current stable.

### DIFF
--- a/install/docker-compose.rst
+++ b/install/docker-compose.rst
@@ -109,7 +109,7 @@ Step 4: Start Zammad using DockerHub images
 
    .. code-block:: sh
 
-      $ docker exec zammad-docker-compose-zammad-railsserver-1 contrib/docker/docker-entrypoint.sh rails r '...your rails command here...'
+      $ docker exec zammad-docker-compose-zammad-railsserver-1 /docker-entrypoint.sh rails r '...your rails command here...'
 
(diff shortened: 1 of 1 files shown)


install/docker-compose.rst
```diff
@@ -88,12 +88,29 @@ Step 4: Start Zammad using DockerHub images
...
+
+   .. code-block:: sh
+
+      $ docker compose run --rm zammad-railsserver rails r '...your rails command here...'
+
+   This will run the command via the docker entrypoint and is recommended. In case you require the use of ``docker exec``, you can use the following command:
+
+   .. code-block:: sh
+
+      $ docker exec zammad-docker-compose-zammad-railsserver-1 contrib/docker/docker-entrypoint.sh rails r '...your rails command here...'
```
MrGeneration, line 112: I'd like to suggest to use `/docker-entrypoint.sh` as entrypoint as this is also the entrypoint used by the Dockerfile and significantly shorter :)

MrGeneration (approved):
Apart from my comment it's looking good to me.

install/docker-compose.rst
mgruner: Good point. Changed accordingly.

PR #348, opened 2023-09-08
Draft: Maintenace - Added Elasticsearch settings (URL/SSL verification)

### LINKED ISSUE #312: PGP: Add GnuPG to the server requirements
```[tasklist]
### Tasks
- [x] Add a section for the new optional dependency `gnupg` to `/prerequisites/software`
- [x] Document new `GPG_PATH` environment variable for optional override of the GnuPG binary path to `/appendix/configure-env-vars`
```

ralf401:


ralf401:
@rolfschmidt I also fixed a TOC issue (roles -> role), you can ignore that ;-)

dvuckovic:
The issue linked to this PR via the commit message is off: it's totally unrelated. But anyway, I think one more task is missing here, see below.

- [ ] Administration via > Console > Query and set / update Zammad settings > Configuring Elasticsearch: Document `es_ssl_verify` boolean flag which will control SSL verification when establishing connection to the server

ralf401:
> The issue linked to this PR via the commit message is off: it's totally unrelated. But anyway, I think one more task is missing here, see below.

Yes because I noticed after pushing the commit that the [Issue](https://github.com/zammad/zammad-admin-documentation/issues/312) is in the wrong repo. 

>     * [ ]  Administration via > Console > Query and set / update Zammad settings > Configuring Elasticsearch: Document `es_ssl_verify` boolean flag which will control SSL verification when establishing connection to the server

You mean to document it also [here](https://docs.zammad.org/en/pre-release/admin/console/hidden-settings.html), right?
Setting would be: `Setting.get('es_ssl_verify')` / `Setting.set('es_ssl_verify', true)`

ralf401:
@dvuckovic I already added the Setting.set('es_ssl_verify', 'true'), see commit https://github.com/zammad/zammad-documentation/pull/348/commits/86f8ef863ff0d5dc78ef62ab84009f70f2050439. Can you please have a look if it's ok? Thank you!

dvuckovic:
> @dvuckovic I already added the Setting.set('es_ssl_verify', 'true'), see commit https://github.com/zammad/zammad-documentation/pull/348/commits/86f8ef863ff0d5dc78ef62ab84009f70f2050439. Can you please have a look if it's ok? Thank you!

The place looks correct to me, but I recommend we reword the comment, to something like `Turn SSL verification on or off`. We should avoid boolean values in human descriptions.

Thanks!

dvuckovic (approved):

PR #362, opened 2023-09-19
Maintenance: Fix styling issues in the documentation builds.

dvuckovic:
The local build can be compared to https://zammad-documentation-dvuckovic.readthedocs.io/en/latest/ in order to spot the differences. They should be minimal now.

In order to influence the local "version" flag, please set the new `ZAMMAD_DOCS_GIT_BRANCH` env variable locally, for example to `main` for `latest` version:

```
export ZAMMAD_DOCS_GIT_BRANCH=main
make html
```

### DIFF
--- a/_static/theme/theme_overrides.css
+++ b/_static/theme/theme_overrides.css
@@ -2,7 +2,7 @@
     max-width: 1050px
 }
 .wy-side-nav-search {
-    background-color: #b95f29;
+    background-color: #19191a;
 }
 
 /* Provide a button and shortened pres to reduce page heights */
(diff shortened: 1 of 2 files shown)


ralf401 (approved):
I noticed the following differences:

- Code-style-text (rendered in red) is sometimes bold in RTD docs (e.g. https://zammad-documentation-dvuckovic.readthedocs.io/en/latest/install/update.html). In the local build, it is not bold at all (which is probably more consequent)
image

- The title of the blocks are highlighted in the RTD build, not in local build (AFAIK you told me that this can't be changed):
image

Apart from that it looks good!!

dvuckovic:
We realized the local builds were not being made in expected way, and after some housekeeping, it proved to output very, very similar builds to RTD.

PR #376, opened 2023-10-16
Fixes #375 - Added CPU cores as an option for rebuilding ES index

### LINKED ISSUE #375: Add rebuild of elasticearch index with multiple cores

ralf401:


ralf401:
@mgruner can you have a look if ok? In my eyes it is not that good but no better position/wording came to my mind. 

Note: I skipped it for the installation of Zammad because there should be not much data to handle.

mgruner (approved):
Looks good to me, thanks @ralf401

PR #378, opened 2023-10-16
Remove sqlite3-related packages from recommended packages

thegcat:
The Zammad `Gemfile` and `Gemfile.lock` don't reference any sqlite-related gems anymore, those packages are probably not needed anymore.


ralf401:
@thegcat thanks again for your contribution!!

ralf401 (approved):

PR #387, opened 2023-10-26
Update other-useful-commands.rst

HanSolo72:
Added description for changing the ticket counter


MrGeneration:
This approach is not officially supported as far as I'm aware and thus at least needs verification by development from my POV.
Also the commands don't seem to be self explaining which might be a huge issue if we add it to the documentation.

ralf401:
Hi @HanSolo72, 

thanks a lot for your PR! I talked about your PR internally and we came to the conclusion that this is not a use case which should occur often / for many users. Sorry, but therefore I close your PR. 

But feel free to contribute to any other issue/topic. :-)

PR #388, opened 2023-11-06
Improve docker compose manual download instructions.

mgruner:
Follow-up for https://github.com/zammad/zammad-docker-compose/issues/382. The manual docker compose download instructions were wrong, as they didn't cover all required files. I now suggest a simpler method that will get all files and also preserve file permissions.


mgruner:
@ralf401 I would like to merge this to `main` right away to make sure it applies immediately.

ralf401 (approved):

PR #399, opened 2023-11-28
Elasticsearch 8 - Add a note for default configuration

AnaelMobilia:
- By default Elasticsearch 8 will use an http**s** connection
- Add commands to define Elasticsearch user / password


MrGeneration:
Just my two cents:
The documentation currently still suggests installing Elasticsearch 7.

In my opinion it should be switched completely to ES8 or, if not, the added statements that only affect ES8 in default scenarios should be pointed out better.

AnaelMobilia:
Thanks @MrGeneration to pointing that.

[As ES7 is still maintained until ES9 release](https://www.elastic.co/support/eol), I updated my change in order to add a note for ES8.

If needed, I could open an issue for updating the doc for ES8 globally.

Regards,
Anael

ralf401:
Hi @AnaelMobilia thanks a lot for your contribution! 

Just 2 minor suggestions:
- replace "encrypted" with "HTTPS" (because the URL is not encrypted itself ;-))
- It would be good to write "Elasticsearch 8+"

It would be nice if you could modify it one more time, then I will merge it. Thank you!

ralf401 (approved):

PR #404, opened 2023-12-08
Update docker-compose.rst for interactive shell

migasQ:
Docu update following https://github.com/zammad/zammad-docker-compose/issues/389 discussion.


mgruner:
Hi @migasQ. Thank you very much for your contribution.
In this case the existing examples should be sufficient. They are only examples for any kind of command you can run in the container, even other than `rails`.

PR #408, opened 2023-12-22
Fix token creation documentaion

mvngne:
I tried to get a token with basic auth, but failed with the given example. The response said, I need to provide "name". Also, after I got the correct json, the response I got was "token", not "name".


MrGeneration:
Thank you for your contribution. The naming in that scope has been changed in... I believe Zammad 6.0 or 6.1.

MrGeneration (approved):

PR #422, opened 2024-01-18
Follow-up: ##338 - Improve SSO related parts slightly

MrGeneration:


appendix/single-sign-on.rst
```diff
@@ -280,50 +287,38 @@ how to reach the *domain controller* (Active Directory server).
                                     (must not be read-only,
                                     but can be the same as ``<domain-controller>``)
 
-.. tabs::
+   .. code-block::
```
ralf401, line 290: In my opinion the code block could be moved out of the note box. But I'm fine with that, too.

ralf401 (approved):
Nice, thanks! Just one comment.

PR #435, opened 2024-02-20
Maintenance: Add support for CentOS 9.

mgruner:
I hope I found all places where CentOS is mentioned / needs to be added.


mgruner (commented):
@ralf401 would you be so kind to check this? You can also feel free to merge it and delete the branch, if it looks ok.

ralf401 (approved):
:+1:

PR #444, opened 2024-03-11
Add an hint for proxy as more configuration is needed

AnaelMobilia:
Put more visible than more configuration is required :-)


ralf401:
Hi @AnaelMobilia thanks a lot for your contribution! I think we don't have a link in all of the prerequisites to the corresponding section. However, it does not hurt so I merge it.

ralf401 (approved):
Thanks a lot for your contribution!

PR #449, opened 2024-03-18
Fixes #446 - Replace commands for unprocessable email handling

### LINKED ISSUE #446: Document handling of failed mails
https://docs.zammad.org/en/latest/admin/console/other-useful-commands.html#reprocess-unprocessable-mails

This command will not work any more with Zammad 6.3 and needs to be removed. Instead, we should document the new rails commands:

```
rake zammad:email_parser:failed_article:reprocess_all  # Reprocess articles which failed to parse
rake zammad:email_parser:failed_email:delete           # Remove a failed email from to the database
rake zammad:email_parser:failed_email:export_all       # Export all failed emails to a local folder
rake zammad:email_parser:failed_email:import           # Import locally modified failed emails back to the database
rake zammad:email_parser:failed_email:reprocess_all    # Reprocess mails which failed to parse
```

ralf401:


admin/console/other-useful-commands.rst
```diff
@@ -13,29 +13,47 @@ This will also show errors that might appear within that process.
...
+
+Edit the Email
+   The email has been exported in the step above. Now you can have a look at it
+   and try to repair it. Make sure to leave the file name untouched, as the
+   import will otherwise fail.
+
+Import Locally Modified Email Back to Database
+   After editing the email, run
+   ``rake zammad:email_parser:failed_email:import /path/to/your/email.eml``
+   to apply your changes from the file to the database.
```
mgruner, line 41: We should mention that you can also pass the entire folder as well, and all .eml files in it will be imported.

admin/console/other-useful-commands.rst
```diff
@@ -13,29 +13,47 @@ This will also show errors that might appear within that process.
...
+   parsing of the imported emails.
+
+   In case Zammad successfully reprocessed emails, you can delete the
+   folder/emails in the file system.
+
+If you don't want to import/reprocess emails, you can delete them from the
+database after exporting them with the command
+``rake zammad:email_parser:failed_email:delete email.eml``. After deleting
+them from the database, you can delete the folder/emails from the file system,
+too.
```
mgruner, line 54: This was changed: the delete command will also delete the file, so we can probably drop the sentence.

admin/console/other-useful-commands.rst
```diff
@@ -13,29 +13,47 @@ This will also show errors that might appear within that process.
 
    >> Channel.fetch
 
-Reprocess unprocessable mails
------------------------------
+Reprocess Unprocessable Emails
```
mgruner, line 16: We should change the naming. Unprocessable mails is not used any more, we now talk about "failed emails" instead.

ralf401: I replaced it in the text as well but kept one "unprocessable" (in brackets) so that it can still be found by searching "unprocessable emails"

mgruner (commented):
Content looks fine @ralf401!

However, I am unsure if this is the proper location, since we are in a section about using the Rails console to do "magic" stuff. Now that we have the commands available, it's a clear interface to use, no longer the Rails console.

What about moving it to the admin manual, email channel documentation? Is the admin manual only for browser based administration, or can it also include calling rake tasks on the commandline?

ralf401:
> However, I am unsure if this is the proper location, since we are in a section about using the Rails console to do "magic" stuff. Now that we have the commands available, it's a clear interface to use, no longer the Rails console.
> 
> What about moving it to the admin manual, email channel documentation? Is the admin manual only for browser based administration, or can it also include calling rake tasks on the commandline?

From my point of view, there is currently no really good place to put it. I'd like to avoid to put it in the admin docs because the idea (so far) was to have more or less only the UI configuration stuff included. So I would leave it for now as it is. A general restructuring of the documentation would be good, for sure. But this should only be done if we switch to a new system.

And most of the readers might not know the difference between the rails console and rake commands.

mgruner (approved):

mgruner:
Ok!

PR #450, opened 2024-03-20
Update kubernetes.rst

AleMScof:
I'd add the create namespace option along with the upgrade command.

### DIFF
--- a/install/kubernetes.rst
+++ b/install/kubernetes.rst
@@ -25,4 +25,4 @@ You need the Helm binary installed / initialized and at least 4 GB of free RAM i
    $ helm repo add zammad https://zammad.github.io/zammad-helm
 
    # Install / Upgrade Zammad
-   $ helm upgrade --install zammad zammad/zammad --namespace=zammad
+   $ helm upgrade --install zammad zammad/zammad --namespace=zammad --create-namespace


ralf401:
@mgruner can you please have a look? Thank you!

mgruner (commented):
@ralf401 thanks a lot. The section on kubernetes in the docs is very short, and also outdated. What do you think about replacing the content with just a link to https://github.com/zammad/zammad-helm/blob/main/zammad/README.md?

ralf401:
> @ralf401 thanks a lot. The section on kubernetes in the docs is very short, and also outdated. What do you think about replacing the content with just a link to https://github.com/zammad/zammad-helm/blob/main/zammad/README.md?

Yes sounds like a good idea if the content is outdated anyway. Another approach would be to port the content from your link to the RTD documentation. What do you think about that?

mgruner:
That readme is part of the helm chart, so it is installed also on helm repo pages like https://artifacthub.io/packages/helm/zammad/zammad. Therefore I'd prefer to maintain the content there.

ralf401:
@AleMScof as discussed we remove the content of this page because it is outdated. So I close your PR.

Anyway, thanks a lot for your contribution!

PR #452, opened 2024-03-21
Fix faulty documentation for renaming of priorities via console.

t-shehab:


ralf401 (approved):
Makes sense, thank you! Do you want to port it to `pre-release` or should I?

PR #459, opened 2024-04-02
Fixes #457 - Replaced  header 'on-behalf' header by 'from' header

### LINKED ISSUE #457: Prefer HTTP standard header "From" over custom "X-On-Behalf-Of" for impersonation
In https://github.com/zammad/zammad/issues/3113, `X-On-Behalf-Of` header was deprecated in favor of `From`. We should update **REST API > Introduction > Actions on behalf of other users** section with the new info.

ralf401:


ralf401:
@dvuckovic not sure if I got it right. Perhaps you can have a look if you have a minute.

PR #468, opened 2024-04-20
Change latest stable version

Dennis1993:
Latest release 5.1 is not correct, the latest version is 6.3

I hope the upgrade path is correct. :)


MrGeneration:
The above text highlights version 5.1 as supposed stable. As in example path to not having to keep that page up to date on every release. If you want to raise it to a more current version, the above text should also reflect that. :)

ralf401:
First, thanks a lot for your contribution. 

However, as @MrGeneration mentioned, this version numbers are only examples and fit to the numbers in the text above. So I close your PR. 

But feel free to contribute again if you find any issues.

PR #529, opened 2024-09-02
Fixes #527 - Changes for package installation

### LINKED ISSUE #527: Adjust OpenSUSE/CentOS install instructions
- Firewall not installed by default
- Wget not installed by default
- Alignment and consistency for different distros, separation of OpenSUSE/SLES where needed
- Locale detection: replace "LC_CTYPE" with LANG
- Add local creation command if there is none

ralf401:


ralf401:
Notes:

- OpenSUSE Leap 15.6 already includes `wget`, so no need to mention additional packages to install
- Added RHEL in tab titles
- Note for installing Zammad before configuring Elasticsearch already present, see [here](https://docs.zammad.org/en/latest/install/elasticsearch.html#step-3-connect-zammad).

t-shehab:
Apart from the comment I posted above, the changes look fine for me (without having tested all the commands, assuming you did).

I'd personally prefer a single note that root privileges are required, and then the commands be given without `sudo`. But that's a matter of opinion, I guess.

PR #532, opened 2024-09-05
Fixes #525 - Reworked update instructions according to comment

### LINKED ISSUE #525: Add "apt-mark hold zammad" to the update suggestion
Because of situations like this:
https://community.zammad.org/t/apt-get-dist-upgrade-y-can-break-your-zammad-install/15344

I suggest changing the update suggestion for debian/ubuntu on this page:
https://docs.zammad.org/en/latest/install/update.html

From this:
```bash
$ apt update
$ apt upgrade
```

To this:
```bash
$ apt update
$ apt-mark hold zammad
$ apt upgrade
$ apt-mark unhold zammad
$ apt upgrade
```

This makes sure Zammad is updated separately after upgrading of other components is done.

ralf401:


ralf401:
Hi @YetAnotherGerrit I updated it based on our call. Please note that I left step 6 untouched because:
- Mentioning Elasticsearch update might make sense if it is running on another system
- Checking release notes might be good at this point to (of course it should have been done at the beginnig) if there are adjustments needed which can be done via Zammad UI for example.

If you disagree, let me know :-D

Download the HTML Zip here => https://github.com/zammad/zammad-documentation/actions/runs/10720051864/artifacts/1896007789

PR #558, opened 2024-11-05
Adds required CSRF token fix for using Caddy with HTTPS

mitchplze:
Tested and working on Zammad 6.3.1. Without this the familiar CSRF error appears when using reverse proxy.

### DIFF
--- a/getting-started/include-csrf-hints.rst
+++ b/getting-started/include-csrf-hints.rst
@@ -21,3 +21,10 @@
 
          RequestHeader set X_FORWARDED_PROTO 'https'
          RequestHeader set X-Forwarded-Ssl on
+
+   caddy
+      Add the following line to your Zammad .env file:
+
+      .. code-block::
+
+         NGINX_SERVER_SCHEME=https    


mitchplze:
I agree. In my opinion it _does_ align with the broad requirements given in the description on the Tip window, though. Caddy is automatically HTTPS out of the box (unless explicitly disabled), which is stated as the goal (of HTTPS) in the tooltip.

image

My PR applies to stock Zammad with stock Caddy put in front of it with the absolute bare minimum Caddyfile config for reverse proxy with HTTPS:

````
examplezammad.com {
  reverse_proxy ipaddress:NGINX_EXPOSE_PORT
}
````

Would it help if I revised with another tab on Caddy, maybe here? I could elaborate and include the above config for example.

image

ralf401:
Hi @mitchplze, thanks a lot for your contribution!

I had a look and if I'm not wrong, your added env variable does the same like the instructions few lines above under _nginx_: it sets the scheme to `https` for nginx. So I would refrain from adding this to the documentation and close this PR.

PR #578, opened 2024-11-25
Fixes #577 - Added note for transaction-based pooling

### LINKED ISSUE #577: Add note about postgres transaction pooling not supported

ralf401:


prerequisites/software.rst
```diff
@@ -162,6 +162,10 @@ You can choose between the following database servers:
 
    .. include:: /appendix/includes/mysql-deprication-note.rst
 
+.. note::
+   Transaction-based connection pooling is not supported.
+   If your setup is built this way, database migrations may fail.
+
```
mgruner, line 168: I think we need to expand this to say that it is for the usage of PgBouncer (see the original issue for details). That seems to be a popular software in cloud environments for Postgresql deployment.
And perhaps we should positively state that instead session connection pooling needs to be used.
See also https://www.pgbouncer.org/features.html

mgruner (changes requested):

prerequisites/software.rst
ralf401: I added PgBouncer as an example, I think there are more solutions. I found [this](https://www.rockdata.net/tutorial/configure-connection-pool/#connection-pooling-solutions) where they mention some of them.


Is it ok now? Feel free to request changes again ;-)

prerequisites/software.rst
```diff
@@ -162,6 +162,12 @@ You can choose between the following database servers:
 
    .. include:: /appendix/includes/mysql-deprication-note.rst
 
+.. note::
+   Make sure to use session connection pooling. Transaction-based connection
+   pooling (e.g. by using PgBouncer) is not supported and may lead to errors
+   during database migrations.
+
+
```
mgruner, line 170: Wording suggestion to make it more clear what it refers to:

If you use database connection pooling software like PgBouncer, make sure to use a pooling mode that is fully compatible with PostgreSQL. Typically this is called "session connection pooling". Transaction-based connection 
 pooling is not supported and may lead to errors during database migrations.

mgruner (approved):
Looks good, but I would still suggest a rephrase because not everyone may understand about what this refers to. Please have a look at my suggestion, and apply it or a similar/improved version. Ok for you @ralf401?

PR #591, opened 2024-12-16
Fixes #590 - Replace APP_RESTART_CMD by AUTO_SHUTDOWN

### LINKED ISSUE #590: Add new Setting "Auto Shutdown" to list of hidden settings
In the next release, we will add a new Setting "Auto Shutdown" that should be explained somewhere in the admin documentation.

WIP: Devs please add details about the Setting-variable here, then you can remove the issue-assignement

```[tasklist]
### Tasks
- [x] Add and describe new setting
- [x] Remove APP_RESTART_CMD (e.g. [here](https://docs.zammad.org/en/latest/appendix/configure-env-vars.html#general-options))
- [x] Check also admin docs if this is mentioned and replace it (see https://github.com/zammad/zammad-admin-documentation/pull/734)
```

ralf401:


appendix/configure-env-vars.rst
```diff
@@ -38,17 +38,19 @@ Use these environment variables to configure Zammad's behavior at runtime.
 General Options
 ---------------
 
-APP_RESTART_CMD
-   The command Zammad will use to automatically restart the server
-   after :admin-docs:`changes have been made in the Object Manager </system/objects.html>`.
-   (*E.g.,* ``"systemctl restart zammad"``)
+AUTO_SHUTDOWN
```
mgruner, line 41: There's a misunderstanding here @ralf401. This is a Zammad setting, not an env var, so it must be set via the Rails console, if required.

ralf401: Oh thanks for the hint! I moved it now to the console section. Hope this is ok now.

mgruner (approved):
Thanks!

PR #595, opened 2025-01-13
Draft: Fixes #584 - Add search endpoint description

### LINKED ISSUE #584: 6.5: Add changes for search endpoint #5384
The following objects have now a unified search endpoint:

Users
Groups
Roles
Organizations
Overviews
Text
Macros
Templates
Tickets
Organizations
Chat Sessions
KnowledgeBase

It looks like there is no documentation at all for the search endpoint. Maybe it makes sense to at least create some generic page for it and it's features.

ralf401:


rolfschmidt (approved):
noice!

MrGeneration:
I'm not the biggest fan to have the search parts within the intro.
Apart from that, the biggest pain point with this change I have is, that the current search documentation is located within the object section it affects. I am aware that this has been drastically improved, but people might search in that place and wonder "where it has gone to".

At least a temporary note might be a good idea.

ralf401:
> I'm not the biggest fan to have the search parts within the intro.
> Apart from that, the biggest pain point with this change I have is, that the current search documentation is located within the object section it affects. I am aware that this has been drastically improved, but people might search in that place and wonder "where it has gone to".
> 
> At least a temporary note might be a good idea.

I would argue that:
- It has to be in the intro because now are some additional endpoints mentioned which are not covered at all
- There is a lot of important information in the intro page, so readers should go through this page anyway
- Many users might come via search engine and land on the most fitting page anyway
- If readers can't find stuff, the search is bad (which indeed is true indeed, but that's one of the reasons we'll have a new system) . But we should not work around a bad search by adjusting / duplicating content.

I'm open for adding a temporary note, but I see no real benefit. Additionally, the docs are getting cluttered and this is not really mainainable.

🚀 as reaction and I put a note in.

PR #606, opened 2025-02-20
Update Ruby version info for 6.5 / Ruby 3.2.7

mgruner:


mgruner:
@ralf401 can you review and perhaps handle this one, please? Are there other places that need an update, such as next.zammad.org? I didn't see a ruby reference there, so it may not be needed.

ralf401 (approved):

PR #615, opened 2025-03-07
Fixes #614 - Adjustments for docker installation

### LINKED ISSUE #614: Adjustment for Docker Compose / Portainer installation
- Move screenshots
- Add to click the "Deploy the stack" button
- Include the env vars in the deployment guide
- Move env vars application box to docker env vars

ralf401:


ralf401:
@mgruner I noticed some things while adding this content to the new docs which I wanted to change. Just some adjustments like moving screenshots and small textual adjustments for the docker stuff so it is now aligned to the new docs.  If you would like to have a look feel free. Otherwise I'll just merge and backport it.

install/docker-compose.rst
```diff
@@ -46,25 +49,22 @@ Step 2: Build From Repository
...
-   If you want to customize the stack, read on in
-   :ref:`the customization section <customizing-stack>` below.
+   In some cases, our default environment is not what a Docker-Compose user is
+   looking for. You can customize the stack using pre-defined scenarios and
+   adjust environment variables. Jump to the
+   :ref:`the customization section <customizing-stack>` below to find more
+   information.
 
-   Zammad runs on port ``8080`` by default. If you want to use another port, you
-   can set it via the variable ``NGINX_EXPOSE_PORT``.
```
mgruner, line 53: Do you think we don't need this any more? (also below)

mgruner (approved):
LGTM. Regarding the NGINX_EXPOSE_PORT I'm not sure, otoh it's also nice to keep the section clean.

install/docker-compose.rst
ralf401: I removed it because it is mentioned in step 3 that Zammad is reachable under port `8080`. I think this is ok for a base deployment. If the reader wants to customize it, the environment variables are mentioned.

But if you think it would be better then we can keep it. What do you think?

mgruner: Yes, should be fine.

PR #628, opened 2025-04-22
Update Ruby requirements for 6.6+

mgruner:


dvuckovic (approved):
@mgruner Looks good, but I also pushed changes for `install/source/include-rvm-install-ruby.rst` snippet as well. Best it's synced with current `develop` state.

mgruner:
Oh, thanks!

PR #631, opened 2025-04-30
Fixes #629 - Added command to disable asciifold

### LINKED ISSUE #629: 6.6: Add ES asciifolding note/info to setup
Maybe we can add some note about the new Elasticsearch asciifolding setting for non-Latin users. So that they can consider, if they want this setting or not.

https://github.com/zammad/zammad/pull/5537

ralf401:


ralf401:
@rolfschmidt I added it to the optional ES settings tab container and linked to a newly created section in the rails console page. Could you have a look if this is ok?

rolfschmidt (approved):

PR #640, opened 2025-05-12
Fixes #597 - Added deb822 method for repository configuration

### LINKED ISSUE #597: Adapt install instructions for Ubuntu/Debian regarding repo key management
https://linuxnews.de/debian-setzt-deb822-fuer-die-quellenliste-um/

ralf401:


ralf401:
@YetAnotherGerrit Can you please have a look? See comment above to download the artifacts.

I applied the following changes:

- Added the de822 method for Elasticsearch installation in addition to the legacy  method.
- Replaced the commands for Ubuntu 24.04 and Debian 12 with the deb822 variant. For both, I placed a hint box which states that we provide the command in the deb822 format starting with these distro versions.

Unrelated to this issue: I separated the code blocks for package installation and Elasticsearch installation to provide better copy/paste experience.

PR #648, opened 2025-05-19
Fixes #536 (CS) - Added Ollama scenario for Docker compose

ralf401:


install/docker-compose/docker-compose-scenarios.rst
```diff
@@ -202,6 +203,26 @@ built in backup service in the stack to save resources.
...
+Add Ollama
+^^^^^^^^^^
+
+You can spin up an additional `Ollama <https://ollama.com/>`_ container to use
+:admin-docs:`Zammad's AI features </ai/features.html>` on your machine.
+
+.. warning:: You should use this only for development or testing purposes and not
+   in production unless you have a good understanding about how to bridge your
+   GPU into a docker container (and even have a GPU in your server) and have a
+   basic understanding about the differences of LLMs and their sizes.
```
mgruner, line 215: Let's reformulate this as a hint instead, and shorten it. We can just positively say that it is intended for testing / development purposes as running a productive LLM stack is a complex matter. Let's not go into the details.

install/docker-compose/docker-compose-scenarios.rst
```diff
@@ -202,6 +203,26 @@ built in backup service in the stack to save resources.
...
+
+To deploy an Ollama container inside the Zammad stack, use the scenario file
+``scenarios/add-ollama.yml``. This creates an Ollama container which
+automatically pulls and serves ``Llama3.2`` to be ready to use/test AI features
+out of the box.
+
+Make sure to use the container's IP or name with the port ``11434`` appended or,
+in the case of a reverse proxy, the URL for the
+:admin-docs:`provider configuration </ai/provider.html>` in Zammad.
```
mgruner, line 224: We know the internal hostname, so why can't we present the final URL `http://ollama:11434` to use for the provider config?
Regarding IP addresses, it's generally not recommendable to use them inside of docker stacks, as they are bound to change on container restart.

mgruner (commented):
Looks nice @ralf401, but can we consider some simplifications please?

install/docker-compose/docker-compose-scenarios.rst
ralf401: Right. I was confused about the name because the container names start with the stack name, e.g. `zammad-docs-ollama-1` and I didn't want to explain such a name structure. But I tested it again, it works when I only add `http://ollama:11434`. I'll change it.

ralf401:
> Looks nice @ralf401, but can we consider some simplifications please?

Thanks for your review! I changed it based on your comments. Feel free to have a look again.

install/docker-compose/docker-compose-scenarios.rst
```diff
@@ -202,6 +203,24 @@ built in backup service in the stack to save resources.
...
+.. hint:: This is intended for development or testing purposes as running a
+   productive LLM stack is complex.
+
+To deploy an Ollama container inside the Zammad stack, use the scenario file
+``scenarios/add-ollama.yml``. This creates an Ollama container which
+automatically pulls and serves ``Llama3.2`` to be ready to use/test AI features
+out of the box.
+
+Add the container name and port (``http://ollama:11434``) or, in case of using
+a reverse proxy, the URL to the
```
mgruner, line 221: Sorry, I forgot to mention this in the first review. What do you mean by reverse proxy here? With the scenario, you get an internal container that can be accessed directly. I don't understand your point about the reverse proxy yet.

install/docker-compose/docker-compose-scenarios.rst
mgruner: Just FYI, there is a difference between the service name (as in the yaml file - this is also used as a hostname inside of the stack, in our case `ollama`) and the container name as it appears in `docker ps`.

ralf401: Ah nice, thanks for clarification! Now it makes sense.

install/docker-compose/docker-compose-scenarios.rst
ralf401: My idea was that when a customer using a reverse proxy, the services could be accessible under different URLs without requiring a port. But I think you are right, it doesn't make sense in this case. Except someone wants to provide the Ollama service for other purposes behind a reverse proxy. I would remove it to avoid confusion.

mgruner (approved):
Thanks a lot @ralf401!

PR #651, opened 2025-05-21
Fixes #647 - Removed MySQL/MariaDB support

### LINKED ISSUE #647: Remove MySQL content and add that only PostgreSQL is supported

ralf401:


ralf401:
@YetAnotherGerrit & @mgruner could you have a look if ok?

I'm not sure if I understood the pgloader comment right in the issue. I kept this for now because users should still be able to migrate from MySQL to PostgreSQL. Maybe  we need a hint about what happens when they upgrade Zammad to version 7 with a MySQL database server?

appendix/migrate-to-postgresql.rst
```diff
@@ -1,10 +1,9 @@
 Migrate to PostgreSQL Server
 ============================
 
-.. include:: /appendix/includes/mysql-deprication-note.rst
-
-The following guide will provide you with a rough direction through that
-migration process.
+The following guide will give you some hints for the migration from
```
mgruner, line 4: What do you think, should we mention that for existing mysql systems they first need to be migrated *before* updating to Zammad 7.0?

prerequisites/software.rst
```diff
@@ -129,36 +129,18 @@ automatically be installed with the Zammad-Package.
...
+`PostgreSQL <https://www.postgresql.org/>`_ 10 or newer.
 
-* PostgreSQL 10+
-* MySQL 5.7+ / MariaDB 10.3+ (⚠️ deprecated with Zammad 7.0+)
-
-.. danger::
-
-   .. include:: /appendix/includes/mysql-deprication-note.rst
+If no PostgreSQL server could be detected, it will be installed automatically
+during the package installation.
```
mgruner, line 136: We refer here to the package installation mode (only). For docker/kubernetes it's different. Is this intended?

mgruner (approved):

prerequisites/software.rst
ralf401: Isn't it only relevant for package installations?

My idea was that a PostgreSQL service is automatically deployed when using Docker/kubernetes (if not deactivated). So they don't have to deal with this requirements.

ralf401:
Not sure if you saw my [comment in the issue](https://github.com/zammad/zammad-documentation/issues/647#issuecomment-2896925314):

> @mgruner should we stick to say that PostgreSQL version 10 or newer is supported? As far as I understand, version 13 is their oldest version which is still in support and updated. And version 17 is the newest.

mgruner:
> Not sure if you saw my [comment in the issue](https://github.com/zammad/zammad-documentation/issues/647#issuecomment-2896925314):
> 
> > @mgruner should we stick to say that PostgreSQL version 10 or newer is supported? As far as I understand, version 13 is their oldest version which is still in support and updated. And version 17 is the newest.

Yes, but let's keep this out of the present MR.

PR #656, opened 2025-06-12
Fixes #570 (CS) - Added new "Backup & Restore (Docker)" page

### LINKED ISSUE #570: Optimize reimplemented Elasticsearch installation instructions
```[tasklist]
### Tasks
- [x] Improve Elasticsearch URL (`https`/`http`) depending on ES versions
- [x] "Routing" of reader either via external docs or to ES7 instructions (note about differences)
```

ralf401:


appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
...
+^^^^^^
+
+By default, a backup is created at each start of the stack as well as at 3
+o'clock each night. The backup is stored in the volume of the
+**zammad-backup** container under ``/var/tmp/zammad``.
+
+Restore
+^^^^^^^
+
+.. warning::
```
mgruner, line 27: This should be no warning. Perhaps just add it to "Copy or move the backup files" below. People can either move all files (then the most recent ones will be used), or they can copy only a specific version of db and app file.

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
...
+   Be aware that the restore process always uses the latest backup according
+   to the timestamp of the file name.
+
+#. Start the new stack at least once so a Zammad database is available.
+#. Copy or move the backup files to ``/var/tmp/zammad/restore/`` inside the
+   volume of the **zammad-backup** container.
+#. Stop the stack.
```
mgruner, line 34: I would recommend to stop the stack before preparing the restore procedure. Why should there be a running stack that is potentially handling incoming mails etc.?

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
...
+``restore`` directory in the docker volume, you can find some examples below.
+
+Restore Inside the Same Stack
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**Requires:** console access to the zammad-backup container.
+
+If you want to restore a backup from the same stack, you just have to create
+the directory and copy/move the files into it. The following example starts the
+**zammad-backup** container and drops the unused one, gets access to the console
```
mgruner, line 55: `drops the unused one, gets access to the console and` - I'm not sure what you mean by this part. This could be deleted.
Nothing gets dropped AFAIR, even if zammad-backup is already running it should keep running.

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
...
+   volume of the **zammad-backup** container.
+#. Stop the stack.
+#. Start the stack. The restore process is triggered if the ``restore``
+   directory is detected and the backup files are in place.
+#. After the restore process has finished, the ``restore`` directory got renamed.
+   You can safely delete it now.
+
+.. _file-handling:
+
+File Handling
```
mgruner, line 42: I wonder if we should move this to a subpage perhaps?

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
+Backup & Restore (Docker)
+=========================
+
+This section shows some basics about the backup and restore process for a docker
+compose based deployment of Zammad. If you already use a different method or
+tool, it is fine to stick with it as long as it works. In such a case, consider
+using the :ref:`disable-backup-service scenario <additional-scenarios>`.
```
mgruner, line 7: + so that the built-in backup and restore mechanism of Zammad is not used.

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,151 @@
...
+
+If you're familiar with docker, the Quick Start section below includes the
```
mgruner, line 9: We should say that the rest of the document describes the built- in backup and restore mechanism.

appendix/backup-and-restore/docker-compose.rst
ralf401: Yeah doesn't really make sense. However, I noticed that the `docker compose cp` command doesn't find the service or container, if the Zammad stack is down. I tried: 

- `docker compose cp /path/of/restore/ zammad-backup:/var/tmp/zammad/`
- `docker compose cp /path/of/restore/ zammad-docker-compose_zammad-backup:/var/tmp/zammad/`

Do you have an idea?

mgruner: Oh, you are right. I was not aware that for `docker cp`, the stack must be running. Sorry.

appendix/backup-and-restore/docker-compose.rst
```diff
@@ -0,0 +1,43 @@
+Backup & Restore (Docker)
+=========================
+
+.. toctree::
+   :hidden:
+
+   docker-file-handling
```
mgruner, line 7: Should we move this to the end of the page, after the Backup and Restore sections?

appendix/backup-and-restore/docker-file-handling.rst
```diff
@@ -0,0 +1,109 @@
+File Handling
```
mgruner, line 1: Maybe we can make the heading more specific. "How to handle restore files"?

ralf401: What do you think about `Backup File Handling`? `How to handle restore files` excludes the fetching of the backup files somehow...

appendix/backup-and-restore/docker-compose.rst
mgruner: If it's too complex, then we can just bring it back to the main document.

But hold on, being a subsection of restore seems to be correct, no?

appendix/backup-and-restore/docker-file-handling.rst
mgruner: Well, it's just for the case when you want to restore something, not about backup files in general…

"How to prepare the restore folder in Docker" or similar?

appendix/backup-and-restore/docker-compose.rst
ralf401: Ok got it to work now with:

`docker compose run --rm -v /path/to/your/host/directory:/restore_tmp zammad-backup bash -c "cp -rv /restore_tmp /var/tmp/zammad/"`

appendix/backup-and-restore/docker-file-handling.rst
ralf401: Same as above applies also here (it is also mentioned how to get the files out of a docker compose deployment). I thought about some variants and now I'm at `File Handling Examples`. The context should be clear as it is a sub-page of the docker backup & restore page.

If you have objections or if you think something else would be better, let me know.

PR #664, opened 2025-06-26
Fixes #663 - Added missing docker environment variables

### LINKED ISSUE #663: Add missing docker environment variables

ralf401:


ralf401:
Sorry I missed the default value for the ES heap size variable. Tried to fix that now.

install/docker-compose/environment.rst
```diff
@@ -103,6 +110,12 @@ Elasticsearch
    * - ELASTICSEARCH_SCHEMA
      - ``http``
      - By default Elasticsearch is reachable via HTTP.
+   * - ELASTICSEARCH_USER
+     - ``elastic``
+     - Set a user to access the Elasticsearch index.
```
mgruner, line 115: I think it's mentioned in the docker-compose.yml file that this value is hardcoded in our built-in container, and cannot be changed. It is only reelvant if an external elasticsearch service is used.
We should either include this info, or not document the setting here.

ralf401: Yeah it is at least mentioned in the documentation somewhere that it is hardcoded. I removed the user and password variable now. Thanks!

PR #703, opened 2025-08-28
Fixes #701 - Unify and fix code blocks and add copy button

### LINKED ISSUE #701: Confusing mix of root and non-root commands
Currently, installation docs contain a confusing mix of shell commands, which should be executed in both root and non-root contexts.

This is OK ✅ 

Image

But, this is not OK ❌ 

Image

Both first and second block look identical, but the second one is misleading: all commands still require root context, even though the `sudo` prefix is missing in them.

Consider introducing the following convention:

1. All non-root command invocations should start just with a `$` shell prefix:

   ```ansi
   $ whoami
   non-root-user
   ```

2. All root command invocations should start either with a `$ sudo` prefix _or_ just with a `#` shell prefix:

   ```ansi
   $ sudo make me a sandwich
   ```

   _or_

   ```ansi
   # make me a sandwich
   ```

   > NOTE: I acknowledge that syntax highlighting may be off when using shell code blocks with `#` prefix, as they are considered comments in shell scripts. However, I'd like to posit that using shell syntax highlighting in such code blocks is just plain _wrong_. These are actually shell **commands**, not a _script_. More correct would be the `ansi` code block or perhaps no syntax highlighting.

ralf401:


admin/console/hidden-settings.rst
```diff
@@ -72,12 +84,12 @@ You can use three different warnings for
...
-.. code-block:: ruby
+.. code-block:: irb
 
    >> Setting.set('ui_ticket_create_notes', {
-         :"phone-in"=>"You're about to note a incoming phone call.",
-         :"phone-out"=>"You're about to note an outgoing phone call.",
-         :"email-out"=>"You're going to send out an E-Mail."
+         "phone-in"=>"You're about to note a incoming phone call.",
+         "phone-out"=>"You're about to note an outgoing phone call.",
+         "email-out"=>"You're going to send out an E-Mail."
```
dvuckovic, line 92: Maybe we can separate the keys from the values a bit, it's less readable this way. So, something like:

```rb
         "phone-in"  => "You're about to note a incoming phone call.",
         "phone-out" => "You're about to note an outgoing phone call.",
         "email-out" => "You're going to send out an E-Mail."
```

admin/console/hidden-settings.rst
```diff
@@ -103,36 +115,36 @@ Adding a Warning to the Article Reply Process
...
-         :"phone-internal" => "You are writing an |internal phone note|, only people of your organization will see it.",
-         :"phone-public"=>"You are writing a |public phone note|.",
-         :"email-internal" => "You are writing an |internal Email|, only people of your organization will see it.",
-         :"email-public"=>"You are writing a |public Email|."
+         "note-internal"   => "You are writing an |internal note|, only people of your organization will see it.",
+         "note-public"     => "You are writing a |public note|.",
+         "phone-internal"  => "You are writing an |internal phone note|, only people of your organization will see it.",
+         "phone-public"    => "You are writing a |public phone note|.",
+         "email-internal"  => "You are writing an |internal Email|, only people of your organization will see it.",
+         "email-public"    => "You are writing a |public Email|."
```
dvuckovic, line 135: `Email` should not be capitalized in English, IMO.

admin/console/hidden-settings.rst
```diff
@@ -103,36 +115,36 @@ Adding a Warning to the Article Reply Process
...
       })
 
 .. note::
 
    You can use example sub-settings above independently, if you e.g. don't need
-   a warning on internal calls, simply leave out ``:"phone-internal"=>""`` out
-   of the setting. The setting itself is in a form of an array ( ``{}`` ).
+   a warning on internal calls, simply leave out ``"phone-internal"`` out
+   of the setting. The setting itself is in a form of an array (``{}``).
```
dvuckovic, line 142: ` ``{}`` ` is a hash, not an array.

admin/console/working-on-tickets.rst
```diff
@@ -13,19 +13,31 @@ check the EML-file itself.
...
+Get article IDs:
+
+.. code-block:: irb
+
+   >> Ticket.find_by(number:'101234').article_ids
+
+Example output:
+
+.. code-block:: irb
+
```
dvuckovic, line 34: For this output code block, we can add `:class: no-copybutton` to its header, it will skip rendering the button.

admin/console.rst
```diff
@@ -94,29 +94,34 @@ It allows you to run several commands inside it.
...
 
-         There was an error trying to connect to Redis via redis://localhost:6379.
-         Please provide a Redis instance at localhost:6379 or set REDIS_URL to point to a different location.
-         ＃<Redis::CannotConnectError: Error connecting to Redis on localhost:6379 (Errno::ECONNREFUSED)>
-         Loading production environment (Rails 6.1.7.3)
-         3.1.3 :001 >
+   **This gives you a response like this:**
 
-      .. code-tab:: console Source Installation
+   .. code-block:: text
```
dvuckovic, line 116: The more correct output type should be `ansi`. But I guess we should also disable the copy button.

install/elasticsearch/troubleshooting.rst
```diff
@@ -56,41 +59,49 @@ Step 1: Verify Elasticsearch is running
...
 
-   .. code-block:: sh
+   .. code-block:: console
 
       $ zammad run rake zammad:searchindex:rebuild[8]
 
    This should start rebuilding the indexes and output it's progress:
 
-   .. code-block:: sh
+   .. code-block:: text
```
dvuckovic, line 104: `ansi` and no copy button?

migration/freshdesk.rst
```diff
@@ -73,177 +72,192 @@ consider using the console over the browser version.
...
+
+         To give you an idea how the migration job state looks like, you can
+         use below tabs. As long as ``finished_at`` is ``nil``, the process is
+         still running.
+
+         .. tabs::
+
+            .. tab:: Freshly started import
+
+               .. code-block:: text
```
dvuckovic, line 138: `ansi` + no copy button?

dvuckovic (changes requested):

admin/console/hidden-settings.rst
ralf401: Oh yeah I missed to add spaces here.

Regarding the E-Mail: I saw it already and left it as it is because there is a GIF below which shows exactly these strings. But when I think about it again, let's fix it in the text at least. GIFs will die at some point anyway.

admin/console.rst
ralf401: Seems this is not known: `WARNING: Pygments lexer name 'ansi' is not known`. I stick with `text` now, unless you have objections.

dvuckovic: `ansi` is probably something from the Vitepress-verse or GitHub-verse, sorry about that. `text` is just fine.

PR #714, opened 2025-09-17
Fixes #708 - Add 'max_connections' function for suggesting a value

### LINKED ISSUE #708: Add rails command for psql max_connections

ralf401:


appendix/configure-database-server.rst
```diff
@@ -39,19 +39,40 @@ Keep in mind that versions may differ from your setup - adapt where needed.
...
 
-   .. code-block:: console
+      As an alternative, these values can be provided as environment variables.
+      In case they are set, the function bypasses the interactive input and just
+      calculates the suggestion. The environment variables are:
 
-      $ sudo systemctl restart postgresql zammad
+      - ``ZAMMAD_MAX_CONNECTIONS_WEB_SERVERS``
+      - ``ZAMMAD_MAX_CONNECTIONS_CONCURRENT_CRONJOBS``
+      - ``ZAMMAD_MAX_CONNECTIONS_CONCURRENT_MANUAL``
```
mgruner, line 62: I think we should skip this part, this is not relevant for everyday usage. This would be interesting only for automation scenarios, which probably don't exist.

mgruner:
@ralf401 looks good in general. I would just simplify this slightly.

ralf401:
Thank you both! I changed it according to your suggestions. Please have a final look if I got it right.

mgruner:
Looks good to me!

Would be nice to have the artifacts regenerated on pushes, not only for the initial PR. Maybe you can create an item for that?

mgruner:
Ah, it seems they may even have been regenerated (article updated).

PR #717, opened 2025-10-02
Draft: Add Debian 13 as supported distro & install instructions

ralf401:
- Added Debian 13 as supported distro
- Added installation instructions for Debian 13
- Changed repo key location to `/usr/share/keyrings/pkgr-zammad-keyring.gpg` for all Debian versions


ralf401:
Closing it because we need to use the new setup to install Debian 13 and have to implement the prerequisites for it first.

PR #744, opened 2025-12-03
Fixes #745 - Add proxy configuration and connection test script

### LINKED ISSUE #745: List of components that allow proxy configuration
Not all gems/components that are used in Zammad respect the proxy settings we have in the admin interface, some purely rely on environment variables or configuration files. To make it more predictable for admins, we should include information how to configure a proxy based on the functionality.

Currently, there are 3 methods, I am aware of, and some components of Zammad require a specific method:
- Our settings: https://admin-docs.zammad.org/en/latest/settings/system/network.html
- Environment variables: `http_proxy`, `https_proxy`, `all_proxy`, `no_proxy` & `ES_JAVA_OPTS`
- Config file (e.g. `/usr/share/elasticsearch/jdk/conf/net.properties`)

ralf401:


rolfschmidt (approved):

ralf401:
@YetAnotherGerrit 

Could you please also have a look? This PR doesn't really solve the [issue](https://github.com/zammad/zammad-documentation/issues/745) (which I moved from admin- to system docs, BTW). 
What do you think, should we keep the issue open? If yes, we should maybe refine and plan it in a sprint, not sure.

YetAnotherGerrit:
The more common and advised way to set a proxy via ENV variables is to use the lower case version `http_proxy` and `https_proxy`. Only specific applications require it to be `HTTP_PROXY` and `HTTPS_PROXY`. If Zammad does not care, I would prefer that we use the lower case version in the documentation. Alternativly add a comment, that it might be interesting to set both.

Ubuntuu e.g. advises for something like this:

```bash
export http_proxy=http://SERVER-IP_ODER_NAME:PORT
export https_proxy=http://SERVER-IP_ODER_NAME:PORT
export ftp_proxy=http://SERVER-IP_ODER_NAME:PORT
export HTTP_PROXY=$http_proxy
export HTTPS_PROXY=$https_proxy
export FTP_PROXY=$ftp_proxy
```

YetAnotherGerrit (approved):

PR #767, opened 2026-02-21
Maintenance: Hint CSRF error fix

dargmuesli:
Source: https://community.zammad.org/t/csrf-token-verification-failed-issue-reverse-proxy-https-setup/17678/3

Finding this took me way too long :wink:

### DIFF
--- a/appendix/environment-variables.rst
+++ b/appendix/environment-variables.rst
@@ -284,6 +284,7 @@ Nginx
      - If the Nginx container for Zammad **is not** the upstream server
        (aka you're using another proxy in front of Nginx) ``$scheme`` may be
        wrong. You can set the correct scheme ``http`` or ``https`` if needed.
+       Set this if you face a ``CSRF Token Verification Failed`` error.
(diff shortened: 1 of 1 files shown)


dargmuesli:
I see, interesting find. I've primarily been looking at the docker documentation for my setup, that's why I think it's reasonable to add a hint to it too. The webserver setup page reads to be aimed at people who actually want to go into configuring a webserver which you'd like to generally "avoid" in the Docker case in which environment variables are the configuration api in that sense. Thus, I think a pointer to the potential CSRF issue fix in the environment variable documentation is very helpful.

---

Ah, speaking about which, I just notice that the page I wanted to edit is this:

- https://docs.zammad.org/en/latest/install/docker-compose/environment.html

But on the current default branch that seems to have been moved to:

- https://docs.zammad.org/en/latest/appendix/configure-env-vars.html

But I think the reason given above still stands.

ralf401:
@dargmuesli thanks for your PR. Looks good to me and I would argue it doesn't hurt to have it there too.

PR #818, opened 2026-04-07
Fix: Resolved broken reference

Daniii44:
Commit 305e8db added an extra step in the a `/appendix/backup-and-restore/restore` which is referenced by `appendix/backup-and-restore/migrate-hosts`.


ralf401 (approved):
Thanks for your contribution!

PR #862, opened 2026-06-23
Document X-Zammad-Suppress-Notifications header for ticket update API

dominikklein:
## Summary

Documents the new `X-Zammad-Suppress-Notifications: true` HTTP header introduced in zammad/zammad#6042.

Adds a tip block to the **Update** section of the Tickets API page explaining:
- How to use the header to suppress agent notifications (email + in-app)
- That it targets automated integrations to avoid notification loops
- Permission scope: agents and admins only, ignored for customers
- That the same header works on `POST /api/v1/ticket_articles` and the GraphQL `ticketUpdate` mutation

## Related

- Implementation MR: https://git.zammad.com/zammad/zammad/-/merge_requests/13394
- GitHub issue: https://github.com/zammad/zammad/issues/6042

### DIFF
--- a/api/ticket/index.rst
+++ b/api/ticket/index.rst
@@ -257,20 +257,19 @@ Required permission: ``ticket.agent`` **or** ``ticket.customer``
 
 .. tip::
 
-   **🔕 Suppress notifications**
+   **Suppress notifications**
 
    To update a ticket without triggering agent notifications (email and
-   in-app), add the following HTTP header to your request::
+   in-app), add the following HTTP header to your request:
+
+   .. code-block:: text
 
       X-Zammad-Suppress-Notifications: true
 
    This is useful for automated integrations that update tickets via
-   webhooks or triggers to avoid notification loops. The header is
-   available to **agents and admins only** — it is silently ignored
-   for customers.
-
-   The header also works on the ``POST /api/v1/ticket_articles``
-   endpoint and the GraphQL ``ticketUpdate`` mutation.
+   webhooks or triggers to avoid notification loops. The header only affects
+   admin and agent accounts and is ignored for customers. It also works for the
+   ``POST /api/v1/ticket_articles`` endpoint.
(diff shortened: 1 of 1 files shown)


dominikklein:
@ralf401 @fliebe92 Maybe you can have a look.

fliebe92 (approved):

api/ticket/index.rst
```diff
@@ -255,6 +255,23 @@ Update
...
+
+      X-Zammad-Suppress-Notifications: true
+
+   This is useful for automated integrations that update tickets via
+   webhooks or triggers to avoid notification loops. The header is
+   available to **agents and admins only** — it is silently ignored
+   for customers.
+
+   The header also works on the ``POST /api/v1/ticket_articles``
+   endpoint and the GraphQL ``ticketUpdate`` mutation.
```
ralf401: Currently, we don't have GraphQL mentioned at all in the docs. This is just related to the new stack, right? Maybe we could omit this part, if it's ok from your side?

api/ticket/index.rst
```diff
@@ -255,6 +255,23 @@ Update
...
+   **🔕 Suppress notifications**
+
+   To update a ticket without triggering agent notifications (email and
+   in-app), add the following HTTP header to your request::
+
+      X-Zammad-Suppress-Notifications: true
+
+   This is useful for automated integrations that update tickets via
+   webhooks or triggers to avoid notification loops. The header is
+   available to **agents and admins only** — it is silently ignored
```
ralf401: I don't really get this. The header can be used by any third-party app/client. But notifications are only suppressed for agents and admins, right? Maybe phrase it like `The header only affects admin and agent accounts and is ignored for customers`? Just an idea.

ralf401 (changes requested):

ralf401 (approved):

PR #863, opened 2026-06-25
Fixes #846 - Adjust supported ES versions and instructions

### LINKED ISSUE #846: Bump up minimum requirement for ES server dependency
With a fix for https://github.com/zammad/zammad/issues/6125 in place, we had to bump up the ES minimum requirement to v7.10. Please update the dependency table at https://docs.zammad.org/en/pre-release/prerequisites/software.html#elasticsearch-optional and check the support across all distributions.

Zammad CI is testing against 7.10.2 patch level from now on.

The target release of the fix is currently Zammad 7.2 so we can announce this in timely fashion, but due to unclear release schedule, it could happen this fix ships in the earlier release (in Zammad 7.1). Therefore, it might be needed to delay the requested documentation change until next release, so the appropriate label is currently set on the issue. Feel free to remove it if the fix is shipped earlier.

ralf401:


appendix/proxy.rst
```diff
@@ -80,7 +79,7 @@ configuration):
    :header: "Address", "Comment"
    :widths: 40, 70
 
-   "artifacts.elastic.co",  "Download of the ingest plugin (only ES < 8)"
+   "artifacts.elastic.co",  "Download of optional plugins"
```
dvuckovic: We don't encourage/hint at any ES plugins anymore, right?!

prerequisites/software.rst
```diff
@@ -217,19 +217,16 @@ system gets.
...
-   "3.4-3.6", ">= 5.5, <= 7.9"
-   "3.3", ">= 2.4, <=7.6"
-   "3.2", ">= 2.4, <=7.5"
-   "3.1", ">= 2.4, <=7.4"
-   "2.0-3.0", ">= 2.4, <=5.6"
-
-An Elasticsearch plugin is required for version 7 or older to index the
-contents of email attachments: ``ingest-attachment``. Starting with
-Elasticsearch 8, it is included by default.
+   "7.2+",        ">= 8.0, < 10"
```
dvuckovic: It's `>=8.11` actually, as previously stated.

From this version the vector DB support was added.

dvuckovic (changes requested):

prerequisites/software.rst
ralf401: Oh sorry, fixed it.

appendix/proxy.rst
ralf401: Yes I removed all the places where plugins (and especially the ingest plugin) were mentioned. Wasn't sure if we should keep it there if someone uses another plugin. But probably fine to remove it because we document a standard installation only anyway.

dvuckovic (approved):

PR #864, opened 2026-06-25
Improve docker restore manual

### LINKED ISSUE #564: Successfull restore do not rebuild the elasticsearch index
**Observed behaviour:**
When you have restored a backup with the "restore-folder-method" all works fine but the elasticsearch-index
have to be rebuilded after the restore. The behaviour now is that you have the restored data but the index is the old.

When the backup was from a near date the gap isn't so big, but as an example when you migrate from an other
system to docker the index might be empty or completely different.

**Expected behaviour:**
_**Idea 1 (needed)**_ : The documentation has to be updated that you have to reindex after restore (Even when you want this)

_**Idea 2 (to discuss)**_: This can be done everytime automaticly in the restore-script that reindexes the restored data in background

_**Idea 3 (to discuss)**_: There can be a restore-folder with complete reindex in parallel to "restore".. like "restore-with-reindex"
(Like the old functionality but when there is the folder ".../restore-with-reindex" the restore reindexes at the end automaticly in background)

mgruner:
Fixes https://github.com/zammad/zammad-docker-compose/issues/564

### DIFF
--- a/appendix/backup-and-restore/docker-compose.rst
+++ b/appendix/backup-and-restore/docker-compose.rst
@@ -52,3 +52,12 @@ Restore
    directory is detected and the backup files are in place.
 #. After the restore process has finished, the ``restore`` directory got renamed.
    You can safely delete it now.
+
+.. hint::
+
+   If you restored into a stack that had data before, you need to perform final clean-ups.
+
+   #. You need to clear the cache of Zammad by running ``Rails.cache.clear``.
+   #. The Elasticsearch index is out of date now, and you should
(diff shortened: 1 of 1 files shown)


mgruner:
@ralf401 can you have a look, please? This belongs to https://github.com/zammad/zammad-docker-compose/issues/564. 
I guess my wording is not optimal, please feel free to change or rewrite it.

mgruner:
@ralf401 maybe this should go straight to `stable`? I backported the changes, but need to do a final docker stack update to get the fixes in.

ralf401 (approved):

PR #870, opened 2026-07-16
Fixes #696 - Overhaul webserver configuration documentation

### LINKED ISSUE #696: Certificate management using ACME
Once available in the nginx packages of our supported distros we should update our documentation to use ACME instead of certbot:

https://www.heise.de/en/news/NGINX-receives-native-ACME-integration-10525873.html

ralf401:


getting-started/configure-webserver.rst
```diff
@@ -3,425 +3,321 @@ Configure the Webserver
...
-
-   The Zammad installation will not automatically set any host- or server name
-   for you.
-
-   | **Docker Compose / Kubernetes users**
-   | Please also note the environment information on
-     :doc:`this page </appendix/environment-variables>`
+You can find sample configuration files for your webserver within the
+``contrib/`` directory of your Zammad installation. During a package
+installation of Zammad, it tries to automatically install a configuration
```
YetAnotherGerrit: I would clarify somehow that this automaticly installed configuration file does not include SSL. Or even, that we provide two example files and in the automatic process the version without SSL is used by default.

getting-started/configure-webserver.rst
```diff
@@ -3,425 +3,321 @@ Configure the Webserver
...
 
          .. include:: includes/nginx-config-paths.include.rst
 
-      Step 2 - Adjust the config file
-         Adjust the just copied file with a text editor of your choice (e.g.
-         ``vi`` or ``nano``).
+      Adjust server name and certificate paths
+         Adjust the just copied file with a text editor of your
+         choice (e.g. ``vi`` or ``nano``).
+         Locate the ``server_name`` directive and adjust
```
YetAnotherGerrit: I would add a hint, that in the SSL-file, the server_name needs to be replaced twice.

YetAnotherGerrit (changes requested):

YetAnotherGerrit (approved):

PR #905, opened 2026-09-08
docs(api): "Removing organizations" → "groups" in group delete note

dominikklein:
The delete note in the Groups API doc said "Removing organizations
with references…", though this endpoint is about groups — a copy from
organization.rst where it is correct. Fixed to "groups".

Scope: 1 line in api/group.rst (organization.rst:249 left as-is).

### DIFF
--- a/api/group.rst
+++ b/api/group.rst
@@ -221,7 +221,7 @@ Required permission: ``admin.group``
 
    Please note that removing groups cannot be undone.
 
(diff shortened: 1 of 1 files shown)


ralf401 (approved):
Thanks!

ralf401:
Ported to `pre-release` with https://github.com/zammad/zammad-documentation/pull/906

PR #912, opened 2026-09-14
Drop Debian 11 from the supported operating systems

fliebe92:
Debian 11 ("bullseye") reached its LTS end of life on 2026-08-31 and no longer receives security updates, so Zammad stops building packages for it.

Changes:

- `prerequisites/software.rst` — Debian 11 removed from the supported operating systems table, and from the `libimlib2` package dependency tab.
- `install/package.rst` — the "Add repository (Debian 11)" block is removed.

The supported operating systems table is included into `install/host-upgrade-repo-migration.rst` via the `supported-os-table-start`/`supported-os-table-end` markers, so that page picks the change up automatically.

⚠️ **Please merge this only after Zammad 7.2 has been released**, so that Debian 11 support is not removed from the documentation too early.

This targets `pre-release` rather than `main` on purpose: Zammad 7.1 keeps receiving Debian 11 packages until it goes out of maintenance, so the current stable documentation has to keep its Debian 11 instructions.

Related: zammad/zammad#6360


ralf401 (approved):
Thanks!

ralf401:
After merge, porting to zammad-org required.

PR #918, opened 2026-09-16
Follow-up #909 - Load Docker Compose scenarios via COMPOSE_FILE instead of include

### LINKED ISSUE #909: Update scenario documentation to use docker-compose.override.yml
As a system administrator, I want the documentation on how to enable scenarios to describe using a `docker-compose.override.yml` file with the `include` keyword, so that I can run scenarios with the plain `docker compose up` command instead of remembering scenario-specific `-f` flag combinations.

## Acceptance criteria

- The docs.zammad.org installation/docker guide no longer instructs admins to run `docker compose -f docker-compose.yml -f scenarios/{scenario}.yml up -d` to enable a scenario.
- The guide instead documents creating a `docker-compose.override.yml` file that references one or more scenario files via `include`, for example:
  - ```yml
    include:
      - scenarios/first.yml
      - scenarios/second.yml

ralf401:


install/docker-compose/docker-compose-scenarios.rst
```diff
@@ -49,39 +49,25 @@ General Usage
...
-    scenario files with additional ``-f`` flags on every command, list the
-    scenarios you want to use in an ``include`` section of a
-    ``docker-compose.override.yml`` file in the cloned repository folder.
-    The stack repository ships an inactive example file you can copy and
-    adjust:
+    scenario files with additional ``-f`` flags on every command, list all
+    compose files in a ``COMPOSE_FILE`` variable in your ``.env`` file in
+    the cloned repository folder (copy ``.env.dist`` if you don't have one
+    yet). ``docker-compose.yml`` always comes first, further files are
+    applied in the order given, and ``:`` separates them:
```
mgruner: Did you take this from the other MR, from what the Agent generated? For me, this seems to be too complex to read and understand.

Image

It's a wall of text with lots of markup. Could you try rephrasing and restructuring this please? We need to assume people for which this is a new topic.

mgruner (changes requested):

install/docker-compose/docker-compose-scenarios.rst
ralf401: Thanks for the suggestion. Tried to make it more simple and clean now. Ok?

install/docker-compose/docker-compose-scenarios.rst
```diff
@@ -47,41 +47,27 @@ General Usage
...
-    .. code-block:: yaml
+    After specifying the scenarios, start the stack with
+    ``docker compose up -d``.
 
-      include:
-        - path: scenarios/{scenario you want to use}.yml
-          project_directory: .
+    .. note::
+       When using the ``COMPOSE_FILE`` variable, the
+       ``docker-compose.override.yml`` is ignored. If you want to use it,
```
mgruner: "is ignored -> is not automatically picked up" maybe?

mgruner (approved):
Thanks! Just one small suggestion still.

install/docker-compose/docker-compose-scenarios.rst
ralf401: Had something like this before and considered "ignored" as more lean. After thinking about it again, "not automatically picked up" is better I would say. I'll add it. Thanks!

PR #923, opened 2026-09-22
Restructure reporting tools pages around single data source workflow

ralf401:
The Grafana sample dashboards pin their index fields and time stamps per panel, so none of the six previously documented data source variants are required. A single Elasticsearch data source using the zammad_production_ index pattern serves all three dashboards.

The Grafana page is reduced to prerequisites, data source setup and the three sample dashboards with screenshots. The dashboard content bullet lists and footnotes are dropped; the screenshots speak for themselves.

The reporting tools landing page is rewritten from scratch: target audience (self-hosted vs. hosted), minimal prerequisites and the tool list. The ignored availability note and the duplicated requirements include are gone, so include-requirements.rst is removed with it.

SaaS customers can enable index access via the Elasticsearch (SaaS) integration on Plus and Ultimate plans; both pages link to it.


ralf401:
@YetAnotherGerrit as announced a first state of the rewrite.

